PR #1, opened 2021-01-26
DRM-1  - create components for Home Page layout (Header, Description, Grid). Create HomePage component. 

OlgaMilczek:
Changes: 
- App.txs - Add StylesProvider injectFirst to fix problem with overwriting styled-components by Material UI. 
- theme.ts - Add breakpoints for , primary color for text, default Font-family. Change font-size for h1 (default 6em). 
Adds: 
- MainContainer.tsx ( Main responsive container with background img, common to all pages)
- Header.tsx (Home page header wit main title)
- Description.tsx
- GridContainer.tsx - container for specialization cards. 
- HomePage.tsx - HomePage main component. (to fill by rest of the content).


src/App.tsx
```diff
@@ -1,14 +1,17 @@
...
-    <ThemeProvider theme={theme}>
-      {/* Replace with actual content */}
-      <DemoContainer />
-    </ThemeProvider>
-  </MuiThemeProvider>
+  <StylesProvider injectFirst>
+    <MuiThemeProvider theme={theme}>
+      <ThemeProvider theme={theme}>
+        {/* Replace with actual content */}
+        <DemoContainer />
```
klaraborowska: You can replace `DemoContainer` with your `HomePageLayout` :)

src/layout/components/Description.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import Typography from '@material-ui/core/Typography';
+import styled from 'styled-components';
+
+const DescriptionContainer = styled('div')`
+  text-align: center;
+  margin: auto;
+  width: 80%;
+  margin: auto;
```
klaraborowska: This style is duplicated (2 lines above).

src/layout/components/Description.tsx
```diff
@@ -0,0 +1,31 @@
...
+        margin-bottom: .8em;
+      }
+  }
+`;
+
+export const Description: React.FC<{}> = () => (
+  <DescriptionContainer>
+    <Typography color="textPrimary">So you want to work in IT? Or you just curious about opportunities in the TECH world?</Typography>
+    <Typography color="textPrimary">You&apos;re in the right place!</Typography>
+    <Typography color="textPrimary">Explore the options below and decide which path works best for you.</Typography>
```
klaraborowska: You could add this style for `color` to body, as we want to have the same primary text color for whole app. This style is inherited by all children, so you could remove it from here than.

src/layout/components/GridContainer.tsx
```diff
@@ -0,0 +1,19 @@
...
+/* Grid container for the main page */
+export const GridContainer = styled('div')`
+    max-width: 960px;
+    width: 80%;
+    display: grid;
+    margin: auto; 
+    margin-top: 2em;
+    grid-template-columns: repeat(auto-fit, minmax(min(350px, 100%), 1fr));
+    grid-column-gap: 2em;
+    grid-row-gap: 2em;
```
klaraborowska: There is a style `grid-gap` which includes both `grid-column-gap` and `grid-row-gap`.

src/layout/components/MainContainer.tsx
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components';
+
+/* Main container with background img for all pages */
+export const MainContainer = styled('div')`
+    background-image: url('./img/background-img.png');
+    background-size: cover; 
+    background-position: center top;
+    background-repeat: no-repeat;
```
klaraborowska: Do you know that there is a shortcut for background styles? Of course your styles are just fine, but if you want you can replace them with `background: url('./img/background-img.png') center top / cover no-repeat;`. The only thing to remember is to separate `background-size` and `background-position` with a slash.

OlgaMilczek: I know :) It is more personal preference, this way is more readable for me (probably because of lack of experience). But maybe I should used to sortcut.

src/layout/components/GridContainer.tsx
OlgaMilczek: Done :)

src/layout/components/Description.tsx
OlgaMilczek: I should add color to index.css or there is some way to do it with a theme? Because I am looking for some solution, but I can't find it.

OlgaMilczek: I find something like this:

```
const theme = createMuiTheme({
  typography: {
    allVariants: {
      color:  '#222222'
    },
  },
});
```

Is this a right solution?

src/App.tsx
OlgaMilczek: Done!

src/layout/components/Description.tsx
OlgaMilczek: Done!

src/layout/components/Description.tsx
klaraborowska: Looks good, if it's working than I think it's fine :) Just check, if it's also working for other elements than material typography, for example `<div>` or `<p>`.  If it's not working  than I would just add this style to `index.css` to `body`.

OlgaMilczek: I check and it actually doesn't work for plain jsx elements, so I add color and font-family to index.css to body and that's work just fine.

klaraborowska (approved):

joannajerz (approved):

src/home-page/components/GridContainer.tsx
```diff
@@ -0,0 +1,18 @@
...
+/* Grid container for the main page */
+export const GridContainer = styled('div')`
+    max-width: 960px;
+    width: 80%;
+    display: grid;
+    margin: auto; 
+    margin-top: 2em;
+    grid-template-columns: repeat(auto-fit, minmax(min(350px, 100%), 1fr));
+    grid-gap: 2em;
+    justify-items: center;
```
vilanelle, line 12: Maybe `justify-items: stretch;` would be a better fit for our needs? :) I don't have much experience with `display: grid;` so feel free to ignore this comment if I'm missing something obvious here :D

src/home-page/components/GridContainer.tsx
```diff
@@ -0,0 +1,18 @@
+import styled from 'styled-components';
```
vilanelle, line 1: Just a suggestion or food for thought: Right now we only use this nice grid container for `HomePage`, but it could be easily reused in other places of the app if needed. What do you think about moving it to a shared folder? But leaving it here for now (as long as we don't reuse it elsewhere) is perfectly fine, too :)

src/index.css
```diff
@@ -1,10 +1,9 @@
 body {
```
vilanelle, line 1: We could move these global styles to styled components: https://styled-components.com/docs/api#createglobalstyle
That way, we'll be able to use our theme easily for global styles. Of course, this is out of scope of this PR - I added a separate task in Jira: https://dareit.atlassian.net/browse/DRM-35

src/App.tsx
```diff
@@ -1,14 +1,17 @@
...
-  <MuiThemeProvider theme={theme}>
-    <ThemeProvider theme={theme}>
-      {/* Replace with actual content */}
-      <DemoContainer />
-    </ThemeProvider>
-  </MuiThemeProvider>
+  <StylesProvider injectFirst>
+    <MuiThemeProvider theme={theme}>
+      <ThemeProvider theme={theme}>
+        {/* Replace with actual content */}
```
vilanelle: I guess we can remove this comment now :)

vilanelle (approved):

src/index.css
OlgaMilczek: Ok. I can do it. :)

src/App.tsx
OlgaMilczek: Ok.

src/home-page/components/GridContainer.tsx
OlgaMilczek: Hmm... I used center because when it is too narrow for two elements it is still too wide for one - in my opinion. But i can add center to media queries for tablets, and for desktop and phone it can be stretch or we can change it when we will see how it works with actual card inside.

PR #2, opened 2021-01-26
DRM-10 Created category card component

Stajgar:
Jira: https://dareit.atlassian.net/browse/DRM-10

Added category card component, ready to be used in the layout page.


.eslintrc.json
```diff
@@ -42,6 +42,7 @@
     "import/prefer-default-export": "off",
     "no-confusing-arrow": "off",
     "arrow-parens": ["warn", "as-needed"],
+    "arrow-body-style": "off",
```
Stajgar: This one gave me some strange alert, which prevented code from compiling. I'll be happy to fix it if anyone knows how to satisfy this requirement in code.

src/App.tsx
```diff
@@ -1,14 +1,23 @@
...
-      {/* Replace with actual content */}
-      <DemoContainer />
-    </ThemeProvider>
-  </MuiThemeProvider>
+  <StylesProvider injectFirst>
+    <MuiThemeProvider theme={theme}>
+      <ThemeProvider theme={theme}>
+        {/* Replace with actual content */}
+        <DemoContainer />
+        <CategoryCard
```
Stajgar: I've added this only to show that it works. It probably should be moved to the layout page. I can remove it from this PR if needed.

Stajgar:
I wasn't sure which directory I should put my component in. For now, I've put it directly in `src`. Maybe we should discuss the file structure and agree on something.

src/App.tsx
klaraborowska: I think you can remove it, beacuse we don't need it for the moment, but it looks nice ;)

src/CategoryCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    padding: theme.spacing(1),
+    maxWidth: 500,
+    margin: '10px auto',
+  },
+}));
+
+type Props = {
```
klaraborowska: Maybe `CardProps` would be a better name, what do you think?

src/CategoryCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const CategoryCard: React.FC<Props> = ({
+  title,
+  description,
+  avatarSrc,
+}) => {
+  const classes = useStyles();
+
+  return (
+    <div className="CategoryCard">
```
klaraborowska: Is this div necessary? I don't see any change after removing it, so maybe it's just some leftover?

.eslintrc.json
klaraborowska: Can you write more about the problem? What is the error msg? It works fine for me without this line of code.

src/CategoryCard.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
klaraborowska: Regarding your question about where should you put this component. I agree, we should definitely discuss the structure. You can start a discussion about it on slack, so all the girls can get involved easily :)

src/CategoryCard.tsx
Stajgar: I haven't got any strong opinion about this. My way of thinking was that these props are already in the category card file and the prefix is self-implying. If it's kind of good practice or convention, let me know, I'll gladly change it.

klaraborowska: I think it's better to give names that are self-explaining. Than looking on the code you know right away, what is this. Otherwise you would have to think in which component you actually are. I wrote you that, because that's exaclty what I did. The project is really small but I came here and I was thinking for few seconds where am I :P

src/App.tsx
Stajgar: Done ✅

src/CategoryCard.tsx
Stajgar: Ok, thanks. It makes sense. I'll rename it ✌️

Stajgar: Done ✅

src/CategoryCard.tsx
Stajgar: Done ✅

joannajerz (approved):

src/CategoryCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+type Props = {
+  title: string;
+  description: string;
+  avatarSrc: string;
+};
+
+export const CategoryCard: React.FC<Props> = ({
+  title,
+  description,
+  avatarSrc,
```
ireshka: Don't we need props for background color of icon? Each card has own color, that can be added to json file with remaining information.

src/CategoryCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <Card className={classes.root}>
+        <CardHeader
+          title={title}
+          subheader={description}
+          action={(
+            <IconButton>
+              <Avatar
+                aria-label="avatar"
+                src={avatarSrc}
+                onClick={() => {
```
ireshka: We discussed about onclick event connected with card. Can we remove this one?

ireshka (changes requested):

src/CategoryCard.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import { Card, CardHeader, IconButton, Avatar } from '@material-ui/core';
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    padding: theme.spacing(1),
+    maxWidth: 500,
```
vilanelle: We should probably also set `minWidth` to make sure Cards are same size :) now, a card with smaller content will shrink:
image

src/CategoryCard.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import { Card, CardHeader, IconButton, Avatar } from '@material-ui/core';
+
+const useStyles = makeStyles((theme) => ({
```
vilanelle: Just wondering: but maybe we want to stick to a single way of customizing components? We need Material UI for components, but for adding our own styles or overriding Material styles we could choose Styled Components. That said,  `useStyles` will work perfectly fine here - I'm suggesting a single styling tool just for the sake of consistency in the app.

src/CategoryCard.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
```
vilanelle: Maybe we could move the Card to another folder? I think we have more or less agreed on the app structure :) I think the appropriate folder is already on `master`.

vilanelle (approved):

ireshka (approved):

src/CategoryCard.tsx
Stajgar: I've tried to fix it. I believe it works, but I'm not sure, so let me know if there's anything I can adjust.

src/CategoryCard.tsx
Stajgar: I've introduced Styled Components to the CategoryCard. It's my first contact with it, so please check out if I applied it correctly.

src/CategoryCard.tsx
Stajgar: Yup, we need it. I've added it ✌️

src/CategoryCard.tsx
vilanelle: Yes, the cards now take up all available space which is the behavior we want here :)

src/home-page/components/CategoryCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const CategoryCard: React.FC<CardProps> = ({
+  title,
+  description,
+  avatarSrc,
+  avatarBackground,
+}) => {
+  return (
+    <CategoryCardContainer
+      onClick={() => {
```
vilanelle: We should probably also add `onClick` to `CardProps` - we will need it to pass the click handler function from the parent component :)

src/CategoryCard.tsx
vilanelle: This is exactly how we should use `styled-components`! :) Working fine!

src/home-page/components/CategoryCard.tsx
Stajgar: Done ✅

.eslintrc.json
```diff
@@ -42,6 +42,7 @@
     "import/prefer-default-export": "off",
     "no-confusing-arrow": "off",
     "arrow-parens": ["warn", "as-needed"],
```
vilanelle, line 44: comment

PR #3, opened 2021-01-28
Drm 12 [Landing page] Routing to Detail Page 

joannajerz:
On card click, we should navigate to respective Detail Page


src/App.tsx
```diff
@@ -1,14 +1,27 @@
...
   <MuiThemeProvider theme={theme}>
     <ThemeProvider theme={theme}>
-      {/* Replace with actual content */}
-      <DemoContainer />
+      <Router>
+        <Switch>
+          <Route exact path="/">
+            <DemoContainer />
+          </Route>
+          <ITRoleRouter path="/backend" proffesion="backend" />
```
joannajerz: only for check if routing is working

tsconfig.json
```diff
@@ -18,9 +18,10 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "noImplicitAny": false,
```
joannajerz: I had a problem with import react-router-dom. I use this to solve the problem

src/common/ITRoleRouter.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { Route } from 'react-router-dom';
+
+interface ITProffesion {
+  proffesion: string;
+  path: string;
+}
+
+export const ITRoleRouter: React.FC<ITProffesion> = ({ proffesion, path }) => (
+  <Route exact path={path} key={proffesion}>
+    <p>Works {proffesion}</p>
```
joannajerz: It will be replace by detail page component

src/App.tsx
```diff
@@ -1,14 +1,27 @@
...
 
 export const App: React.FC<{}> = () => (
   <MuiThemeProvider theme={theme}>
     <ThemeProvider theme={theme}>
-      {/* Replace with actual content */}
-      <DemoContainer />
+      <Router>
+        <Switch>
+          <Route exact path="/">
+            <DemoContainer />
```
joannajerz: It will be replace by homepage component

tsconfig.json
ireshka: I wonder if we can solve this issue another way :) Probably we will need this eslint rule set to true for better typescript code.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
vilanelle: I think ide settings are typically excluded from version control - we could add this file to `.gitignore`?

yarn.lock
```diff
@@ -1141,7 +1141,7 @@
   dependencies:
```
vilanelle: I'm afraid we need to stick to one package manager for the project, so can we please remove this file and use `npm install`? If we have both `yarn-lock` and `package-lock.json`, we risk unreliable dependency management.

tsconfig.json
vilanelle: There is - we need to install manually types for react-router package :) 

`npm i --save-dev @types/react-router-dom`

image

This is actually going to happen quite often that we will have to install such types ourselves when adding new dependencies.

src/common/ITRoleRouter.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Route } from 'react-router-dom';
+
+interface ITProffesion {
+  proffesion: string;
```
vilanelle: Sorry to complain about such a small detail, but could we please check the spelling? I think it's 'profession'. I know it seems like a very small thing, but it can be a bit of a problem when searching for components.
I usually work on Webstorm, which catches any typos for me, for VS Code my colleagues recommend this plugin to do the same :)
https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

tsconfig.json
joannajerz: ok thanks, I found another solution;)

.vscode/settings.json
joannajerz: done

src/common/ITRoleRouter.tsx
```diff
@@ -0,0 +1,13 @@
...
+  path: string;
+}
+
+export const ITRoleRouter: React.FC<ITProffesion> = ({ proffesion, path }) => (
+  <Route exact path={path} key={proffesion}>
```
vilanelle: Just wondering: are going to have a situation where `path` and `profession` props will be different? Maybe one prop would be enough - what do you think?

vilanelle (approved):

src/ITProfessions.ts
```diff
@@ -0,0 +1,7 @@
+export const ITProfessions = [
```
klaraborowska, line 1: I'm not sure, if it's the best place for this file. I would rather move to some common folder, as we may want to use it in other places as well. But we can leave it for now here and decide later about it.

src/ITProfessions.ts
```diff
@@ -0,0 +1,7 @@
...
+  'Backend',
```
klaraborowska, line 2: I'm not sure, if it's the best place for this file. I would rather move to some common folder, as we may want to use it in other places as well. But we can leave it for now here and decide later about it.

klaraborowska (approved):

PR #4, opened 2021-02-04
Drm 11 - Add a file with data and display cards with the relevant inputs

ireshka:
Except main functionalitis I added some code format changes because of eslint warnings.


src/home-page/HomePage.tsx
```diff
@@ -4,13 +4,28 @@ import { MainContainer } from '../layout/components/MainContainer';
...
+      <GridContainer>
+        {cards.map(({ title, description, color, icon }) => {
+          return (
+            <CategoryCard
+              key={title}
+              title={title}
+              description={description}
+              avatarSrc={icon}
+              avatarBackground={color}
+              clickHandler={() => { return true; }}
```
vilanelle: We probably added this return value, because the linter complained about `any` in the `CardProps` typings. What we could do instead is change it to: 
```
 clickHandler={() => null}
```
And in `CategoryCard`:
```
type CardProps = {
  title: string;	
  ...
  clickHandler: () => void;
};

File: src/home-page/components/CategoryCard.tsx [hunk 2]
```diff
@@ -7,26 +7,27 @@ const CategoryCardContainer = styled(Card)`
...
+  width: 100%;
 `;
 
-const AvatarContainer = styled(Avatar)<{backgroundColor: string}>`
- margin-right: 1em;
- background-color: ${props => props.backgroundColor};
+const AvatarContainer = styled(Avatar)<{ background: string }>`
+  margin-right: 1em;
+  background-color: ${({ background }) => { return background; }};
+  padding: .75rem;
```
vilanelle: Is there any particular reason we need to have `rem` for padding, when we have `em` for margin? If not, I'd suggest we choose one and stick with it :)

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
vilanelle: Hm, I don't think we have prettier installed for our app? :) The settings below can be (and some already are) configured using `eslint`.

src/icons/backend.svg
```diff
@@ -0,0 +1 @@
...="19" cy="11" r="1"/><path fill="#ffffff" d="M24,24a4,4,0,1,0,4,4A4,4,0,0,0,24,24Zm0,6a2,2,0,1,1,2-2A2,2,0,0,1,24,30Z"/><path fill="#ffffff" d="M33.66,24.73l-2-3.46a1,1,0,0,0-.95-.5l-2,.17a7.86,7.86,0,0,0-1-.56l-.84-1.81A1,1,0,0,0,26,18H22a1,1,0,0,0-.91.57l-.84,1.81a7.86,7.86,0,0,0-1,.56l-2-.17a1,1,0,0,0-.95.5l-2,3.46a1,1,0,0,0,0,1.07l1.14,1.64c0,.18,0,.37,0,.56s0,.38,0,.56L14.38,30.2a1,1,0,0,0,0,1.07l2,3.46a1,1,0,0,0,.95.5l2-.17a7.86,7.86,0,0,0,1,.56l.84,1.81A1,1,0,0,0,22,38h4a1,1,0,0,0,.91-.57l.84-1.81a7.86,7.86,0,0,0,1-.56l2,.17a1,1,0,0,0,.95-.5l2-3.46a1,1,0,0,0,0-1.07l-1.14-1.64c0-.18,0-.37,0-.56s0-.38,0-.56l1.14-1.64A1,1,0,0,0,33.66,24.73ZM30.5,28c0,.25,0,.49,0,.73a1,1,0,0,0,.17.68l1,1.41-1.36,2.36L28.53,33a1,1,0,0,0-.68.2A6.9,6.9,0,0,1,26.6,34a1,1,0,0,0-.51.49L25.36,36H22.64l-.73-1.55A1,1,0,0,0,21.4,34a6.9,6.9,0,0,1-1.25-.73,1,1,0,0,0-.68-.2l-1.72.15-1.36-2.36,1-1.41a1,1,0,0,0,.17-.68c0-.24,0-.48,0-.73s0-.49,0-.73a1,1,0,0,0-.17-.68l-1-1.41,1.36-2.36,1.72.15a1,1,0,0,0,.68-.2A6.9,6.9,0,0,1,21.4,22a1,1,0,0,0,.51-.49L22.63,20h2.74l.72,1.55a1,1,0,0,0,.51.49,6.9,6.9,0,0,1,1.25.73,1,1,0,0,0,.68.2l1.72-.15,1.36,2.36-1,1.41a1,1,0,0,0-.17.68C30.48,27.51,30.5,27.75,30.5,28Z"/></g></svg>
```
vilanelle, line 1: Great job with formatting the icons 😍
image

src/home-page/data/data.ts
```diff
@@ -0,0 +1,50 @@
+import backendIcon from '../../icons/backend.svg';
+import devopsIcon from '../../icons/devops.svg';
+import frontendIcon from '../../icons/frontend.svg';
+import projectIcon from '../../icons/project_manager.svg';
+import qaIcon from '../../icons/qa.svg';
+import uxIcon from '../../icons/ux_ui.svg';
+
+export const cards = [
```
vilanelle: We have a really nice separation of state and logic with all the data extracted to this file 🎉However, we should have a single source of truth for our IT professions list. What I mean, is that we already have `ITProfessions.ts` file and that gives us two sources of truth that can potentially diverge and lead to inconsistencies and errors in the future. 
Perhaps we could merge these two sources? To do it we would remove the `ITProfessions.ts` file and add a 'route' prop to data, like so: 
```
{
    id: 'frontend',
    title: 'Frontend',
    description: 'A frontend engineer works on those parts of web apps you actually see',
    color: '#bee5e4',
    icon: frontendIcon,
    route: 'frontend'
  },
```
I'd also add `id` prop and rename the `const` to `ITProfessions` from `cards` so that the name reflects what data we hold here rather than their function. We could then use the new `const` in `App` component to map out the routes.
Finally, I'd consider renaming the file (to make the file name the same as `const` name) and moving the data folder up to `src` directory - after the changes it will be relevant to the whole app rather than only `Home Page`.
This is quite a lengthy comment, so do let me know if not everything is clear 😅

src/home-page/components/GridContainer.tsx
```diff
@@ -5,13 +5,13 @@ export const GridContainer = styled('div')`
...
-    margin: auto; 
-    margin-top: 2em;
+    margin: 2em auto 0;
+    padding-bottom: 2em;
     grid-template-columns: repeat(auto-fit, minmax(min(350px, 100%), 1fr));
     grid-gap: 2em;
     justify-items: center;
 
-    ${props => props.theme.breakpoints.down('sm')} {
+    ${props => { return props.theme.breakpoints.down('sm'); }} {
```
vilanelle: Hm, I see I missed a thing in our lint config.. :/ we don't really want to include braces and explicit `return` for return-only arrow functions, because one of the main advantages of arrow functions is that they are so nice and short :) This is probably the lint configuration we want ([documentation examples here]( https://eslint.org/docs/rules/arrow-body-style)):
```
"arrow-body-style": ["warn", "as-needed"],
```
Good news is that we don't have to change the code back manually! After changing the eslint rule, we simply need to run `npm run lint-fix` and everything will correct itself :)

vilanelle (changes requested):

src/home-page/data/data.ts
```diff
@@ -0,0 +1,50 @@
+import backendIcon from '../../icons/backend.svg';
+import devopsIcon from '../../icons/devops.svg';
+import frontendIcon from '../../icons/frontend.svg';
+import projectIcon from '../../icons/project_manager.svg';
+import qaIcon from '../../icons/qa.svg';
+import uxIcon from '../../icons/ux_ui.svg';
```
OlgaMilczek: Mayby we can use one file with all svg icons together? To limit imports. What do you think? This solution is also fine of course, and we can stay with that :)

src/home-page/HomePage.tsx
```diff
@@ -4,13 +4,28 @@ import { MainContainer } from '../layout/components/MainContainer';
...
-    </GridContainer>
-  </MainContainer>
-);
+export const HomePage: React.FC<{}> = () => {
+  return (
+    <MainContainer>
+      <Header />
+      <Description />
+      <GridContainer>
+        {cards.map(({ title, description, color, icon }) => {
```
OlgaMilczek: We can add some comment here to remembered what is going on in the future. Becouse I think name "cards" is not self-documenting. Also we can think about changing name cards for something more self-explanatory eg. itProfesionsCards to make it more clear.

src/home-page/data/data.ts
```diff
@@ -0,0 +1,50 @@
...
+    description:
+      'A DevOps creates and maintains infrastructure that software runs on',
+    color: '#d99cff',
+    icon: devopsIcon,
+  },
+  {
+    title: 'Project Manager',
+    description:
+      'Project managers plan, monitor, control and organize projects',
+    color: '#aafa87',
```
OlgaMilczek: We can add this colors to theme because we will like to reuse it over the page. 
https://material-ui.com/customization/palette/

src/home-page/components/CategoryCard.tsx
```diff
@@ -37,14 +38,9 @@ export const CategoryCard: React.FC<CardProps> = ({
...
-      onClick={clickHandler}
-    >
-      <CardHeaderContainer
-        title={title}
-        subheader={description}
-      />
-      <AvatarContainer src={avatarSrc} backgroundColor={avatarBackground} />
+    <CategoryCardContainer onClick={clickHandler}>
+      <CardHeaderContainer title={title} subheader={description} />
+      <AvatarContainer src={avatarSrc} background={avatarBackground} />
```
klaraborowska: I think you could add redirection to the details screen already, casue the rooting is ready.

src/home-page/data/data.ts
klaraborowska: @OlgaMilczek to be honest I had no idea that you can have multiple svg in one file and use them separately, so a good lesson for me today :) (but I'm not sure, if it was easy/possible to implement in this particular case). Anyway, I would leave it as it is. It's clear and easy to read or update the fille in case you want to change one of the icons in the future.

OlgaMilczek: @klaraborowska Yes you are right I confused import svgs in html and js. :( My bad. @ireshka Sorry I is perfect :)

vilanelle: We could make these `svgs` react components rather than just svg files and then we would have more options concerning imports / exports :) But this is just a note, I agree that the current solution is just fine :ok_hand:

.prettierrc
ireshka: I added prettierrc config that is compatible with eslint for easy formatting code in VSC (it uses .prettierrc config file for format). Should I add it to .gitignore if I'd like to use it?

src/home-page/HomePage.tsx
ireshka: I resolved it by adding redirection handler as @klaraborowska mentioned below.

src/home-page/components/CategoryCard.tsx
ireshka: It was just a mistake. I corrected it and use `em` unit.

src/home-page/data/data.ts
ireshka: You are absolutely right. I created one file with required data as you pointed.

src/home-page/components/GridContainer.tsx
ireshka: Corrected.

src/home-page/HomePage.tsx
ireshka: I changed filename for ITProfessions as @vilanelle suggested above.

src/home-page/data/data.ts
ireshka: Good idea. I modifed our palette and added colors.

src/home-page/components/CategoryCard.tsx
ireshka: Done.

.prettierrc
vilanelle: Exactly! :) If it's a dependency that you're using locally, it should go to `.gitgnore`. Alternatively, if we want to add `prettier` to our project, we should install it as a dev dependency.

src/layout/theme.ts
```diff
@@ -5,6 +5,31 @@ import green from '@material-ui/core/colors/green';
 
 const breakpoints = createBreakpoints({});
 
+declare module '@material-ui/core/styles/createPalette' {
```
vilanelle, line 8: Nice! :)

src/home-page/components/CategoryCard.tsx
```diff
@@ -7,26 +7,28 @@ const CategoryCardContainer = styled(Card)`
...
- width: 100%;
+  width: 100%;
 `;
 
-const AvatarContainer = styled(Avatar)<{backgroundColor: string}>`
- margin-right: 1em;
- background-color: ${props => props.backgroundColor};
+const AvatarContainer = styled(Avatar)<{ background: string }>`
+  margin-right: 1em;
+  background-color: ${({ background, theme }) => theme.palette.professions[background].main};
```
vilanelle, line 22: We're accessing `background` prop here with bracket notation which is convenient, but unfortunately also means we are no longer guarded by TypeScript type checking. What I mean is that if we pass a wrong string to `AvatarContainer`, e.g. we make a typo and pass `fontend`, we may break our code.
What we could do to is use a type union: https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html. To do this, we need to declare the following:

`type ProfessionKey = 'frontend' | 'backend' | 'uxui' | ... ;`

And add an interface for props:
```
interface Props = {
   background: ProfessionKey;
}
```

vilanelle (approved):

klaraborowska (approved):

PR #5, opened 2021-02-06
DRM-17 - Added detail page routing and component with dummy data

Stajgar:
Jira: https://dareit.atlassian.net/browse/DRM-17


src/App.tsx
```diff
@@ -20,9 +21,9 @@ export const App: React.FC<{}> = () => (
             <Route exact path="/">
               <HomePage />
             </Route>
-            {ITProfessions.map((r) => (
-              <Route exact path={`/${r}`} key={r}>
-                <p>Works {r}</p>
+            {ITProfessions.map((profession) => (
+              <Route exact path={`/details/${profession.name}`} key={profession.name}>
```
vilanelle: I like it a lot how we have a common sub-route for all professions (`/details`)! 👍

src/ITProfessions.ts
```diff
@@ -1,7 +1,22 @@
...
+  },
+  {
+    name: 'Devops',
+    description: 'Devops is awesome',
+  },
+  {
+    name: 'Project-Manager',
+    description: 'Project-Manager is awesome',
+  },
+];
```
vilanelle: It's great that we are keeping our state (professions data) in one place, meaning we have a single source of truth for the app! 
I'd also go a step further and added another prop: `route`. At the moment, we are using `name` as path for our routes and that may cause some issues. For example, we have to add a hyphen in `Project-Manager` to be able to use it for route, but we probably don't want the hyphen if use the `name` for page header. 
We could also add `id` prop that would not be subject to change (`name` may easily change and we want to have some constant thing to track professions by). So in the end we could have something like this:
```
{
    id: 'uxUi',
    name: 'UX-UI',
    description: 'UX-UI is awesome',
    route: 'ux-ui',
},
```

src/ITProfessions.ts
```diff
@@ -1,7 +1,22 @@
-export const ITProfessions = [
-  'Backend',
-  'Frontend',
-  'UX-UI',
-  'Devops',
-  'QA',
-  'Project-Manager'];
+export const ITProfessions: { name: string, description: string }[] = [
```
vilanelle: It's great that we have type checking here 👏 Maybe we could also extract it to an interface, like so:
```
interface ITProfession {
   name: string;
   description: string;
}
```
and: 
```
export const ITProfessions: ITProfession[] = ...
```
At present, we are inlining it and that's totally fine, but if our interface grows (see my other comment) a separate interface may be more readable :)

vilanelle (changes requested):

src/ITProfessions.ts
vilanelle: I actually suggested similar changes concerning our state / data to @ireshka in her PR: https://github.com/vilanelle/tech-transition/pull/4#discussion_r571227394. There I described how we would like our state to look like in the end. If you wish, you can apply it here or you can just do the smaller change (as described above) for the scope of this PR, I just wanted to give you a heads up :)

src/App.tsx
```diff
@@ -20,9 +21,9 @@ export const App: React.FC<{}> = () => (
...
+                <DetailPage profession={profession.name} description={profession.description} />
```
klaraborowska: We will need also other information inside details component (for example icon or color for each path), so I would suggest to pass the whole profession object as props.

src/ITProfessions.ts
Stajgar: Done

src/App.tsx
Stajgar: Done

src/ITProfessions.ts
Stajgar: I've change everything as you suggested. As for addressing changes in Gosia's PR, I think the last of us who'll be merging PR should unify the changes.

src/App.tsx
```diff
@@ -20,9 +21,9 @@ export const App: React.FC<{}> = () => (
             <Route exact path="/">
               <HomePage />
             </Route>
-            {ITProfessions.map((r) => (
-              <Route exact path={`/${r}`} key={r}>
-                <p>Works {r}</p>
+            {ITProfessions.map((profession) => (
+              <Route exact path={`/details/${profession.route}`} key={profession.route}>
```
vilanelle: Typically, we'd use `id` as `key`, because it's should be a unique value - although in this case `route` also meets this requirement.

vilanelle (approved):

klaraborowska (approved):

src/App.tsx
Stajgar: Ok, I've changed it

PR #6, opened 2021-02-08
Drm 19 - Add tabs container for detail page

OlgaMilczek:
Changes:
- theme.ts - Add primary color to palette, Add disable text color, text color to palette

Adds:
- StyledTab.tsx component for creating one tab heading
- TabCategiores.tsx - maybe this file content should be add to App state and passing in props?
- TabPanel - component for creating one Panel with tab content. 
- TabContainer - Main component, which create Tabs.
- JobTab, LearnTab, ResourcesTab empty components need to be fill with content with next sprint.


src/detail-page/components/TabCategories.tsx
```diff
@@ -0,0 +1,19 @@
...
+  },
+  {
+    name: 'jobs',
+    text: 'Jobs',
+  },
+  {
+    name: 'resources',
+    text: 'Useful Resources',
+  },
+];
```
OlgaMilczek: Maybe this file content should be add to App state and passing in props?

src/detail-page/components/ResourcesTab.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+interface TabProps {
+    content: any;
```
vilanelle: I know it's a very small thing, but maybe we could unify indentation a bit? In several files we have varying indentation e.g. here in the interface we have 4 spaces, while in component below it's 2 spaces.

src/detail-page/components/StyledTab.tsx
```diff
@@ -0,0 +1,30 @@
...
+`;
+
+interface TabProps {
+    text: string;
+    chosenTab: number;
+    index: number;
+    onClick: (index: number) => any;
+  }
+
+export const StyledTab: React.FC<TabProps> = ({ text, chosenTab, index, onClick }) => {
```
vilanelle: I see that we have a problem with duplicate names here. What we can do, is rename the import:

`import { Tab as MuiTab } from '@material-ui/core';`

Then, we can name our component simply `Tab` :)

src/detail-page/components/JobsTab.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+interface TabProps {
```
klaraborowska: I think it would be nice to rename it to something like JobsProps (and in other tabs accordingly so LearnProps etc.)

src/detail-page/components/StyledTab.tsx
```diff
@@ -0,0 +1,30 @@
...
+  /* Adding className active to chosen card for creating bold text */
+  return (
+    <TabWithStyle
+      label={text}
+      className={chosenTab === index ? 'active' : ''}
+      onClick={() => { return onClick(index); }}
```
klaraborowska: `onClick={() => onClick(index)}` would be enough here.

src/detail-page/components/TabsContainer.tsx
```diff
@@ -0,0 +1,82 @@
...
+      tabContent = <JobsTab content={jobs} />;
+      break;
+    } case 'learn': {
+      tabContent = <LearnTab content={learn} />;
+      break;
+    } case 'resources': {
+      tabContent = <ResourcesTab content={resources} />;
+      break;
+    } default: throw new Error('Wrong category name!');
+    }
```
klaraborowska: You could work on formatting this switch statement as it's a bit difficult to read it, here nice example https://developer.mozilla.org/pl/docs/Web/JavaScript/Referencje/Polecenia/switch

src/detail-page/components/StyledTab.tsx
```diff
@@ -0,0 +1,30 @@
...
+    &.active {
+        font-weight: bold;
+    }
+`;
+
+interface TabProps {
+    text: string;
+    chosenTab: number;
+    index: number;
+    onClick: (index: number) => any;
```
klaraborowska: This onClick functon will only set selected tab index, so it will not return anything. We could change `any` to `void`.

src/detail-page/components/TabCategories.tsx
klaraborowska: We should pass it as props, depending on which profession we select.  We may want to have different tabs (or more/less tabs) for some professions.

src/detail-page/components/TabsContainer.tsx
klaraborowska: Thinking about it more, it would be tempting  for me  to do tabs with routing. It would be a nice practice, but it's not a requirement, so I will leave it up to you.

src/detail-page/components/TabsContainer.tsx
```diff
@@ -0,0 +1,82 @@
...
+    jobs: unknown;
+    resources: unknown;
+  }
+
+export const TabsContainer: React.FC<TabProps> = ({
+  learn,
+  jobs,
+  resources,
+}) => {
+  const [choosesCategory, setChoosesCategory] = useState(0);
```
klaraborowska: Maybe chosenCategory and setChosenCategory?

src/detail-page/components/TabPanel.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+interface TabPanelProps {
+    children?: React.ReactNode;
+    index: unknown;
+    value: unknown;
```
klaraborowska: Why is it unnknown? It looks like it should be a number for index and value.

src/detail-page/components/TabCategories.tsx
vilanelle: That's a very good point! However, we currently have quite a lot of changes going on in the app state (I mean PRs of Gosia and Agnieszka), so maybe we can leave it as is for now, to avoid more conflicts :) But we should probably review app state by the end of this sprint and decide how we want to refactor :) We may even consider if we want to add some state management tools ☝️

vilanelle (changes requested):

klaraborowska (changes requested):

src/detail-page/components/StyledTab.tsx
vilanelle: I guess this was enforced by our eslint rule - but it's been changed in Gosia's PR, so we can follow Klara's suggestion :)

src/detail-page/components/TabsContainer.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const TabsHeading = categories.map((category: any, id: number) => {
```
vilanelle: We shouldn't set `category` type as `any` - it's actually `Category` :) we can set it if we export `Category` type from `TabCategories.tsx`. But we actually don't have to set a type by ourselves - TypeScript is clever enough to infer the type from the value we're passing (`categories` array) :) https://www.typescriptlang.org/docs/handbook/type-inference.html
On an unrelated issue: we should rename `id` to `index` to reflect more accurately what this variable is.

src/detail-page/components/TabsContainer.tsx
```diff
@@ -0,0 +1,82 @@
...
+import styled from 'styled-components';
+
+import { TabPanel } from './TabPanel';
+import { StyledTab } from './StyledTab';
+import { JobsTab } from './JobsTab';
+import { LearnTab } from './LearnTab';
+import { ResourcesTab } from './ResourcesTab';
+import { categories } from './TabCategories';
+
+const StyledTabs = styled(Tabs)`
```
vilanelle: Also here we can rename import ( `import { Tabs as MuiTabs } from '@material-ui/core';`) and then our component can be named simply `Tabs`! :)

src/detail-page/components/StyledTab.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
vilanelle: What do you think about adding a new folder inside `detail-page/components` directory that would contain out tab components (without tab content components)? :)
image

src/detail-page/components/TabsContainer.tsx
```diff
@@ -0,0 +1,82 @@
...
+    );
+  });
+
+  return (
+    <StyledCard>
+      <StyledTabs
+        indicatorColor="primary"
+        value={choosesCategory}
+      >
+        {TabsHeading}
```
vilanelle: I'm not sure if we should combine lists (such as `categories.map(..)`) in collective components such as `TabsHeading` or `TabPanels`. Let's make use of React's ability to map inside out templates - that will give us better visibility of what is actually happening inside the component.
For `TabsHeading` all we have to do is replace it with `categories.map((..) => <StyledTab .../>)`. For `TabPanels` we can do something like this:
```
 {categories.map((category, id) => (
        <TabPanel value={choosesCategory} index={id} key={category.name}>
          {getTabContent(category.name)}
        </TabPanel>
      ))}
```
Where `getTabContent` would be the `switch`, refactored to return React elements 🙃 
```
const getTabContent = (name: string) => {
    switch (name) {
    case 'jobs': {
      return <JobsTab content={jobs} />;
    }

**Review by vilanelle — CHANGES_REQUESTED** [review]

File: src/detail-page/components/StyledTab.tsx (hunk 5 above)
Comment (klaraborowska) [reply src/detail-page/components/StyledTab.tsx]: Sorry Olga , I didn't notice it before. Ania you're right, I see it now also in other pr. I think we can fix it later  in the whole project :)

File: src/detail-page/components/ResourcesTab.tsx (hunk 2 above)
Comment (OlgaMilczek) [reply src/detail-page/components/ResourcesTab.tsx]: Ok. Done. I check all my files I hope It will be ok now.

File: src/detail-page/components/StyledTab.tsx (hunk 3 above)
Comment (OlgaMilczek) [reply src/detail-page/components/StyledTab.tsx]: Done

File: src/detail-page/components/JobsTab.tsx (hunk 4 above)
Comment (OlgaMilczek) [reply src/detail-page/components/JobsTab.tsx]: Done!

File: src/detail-page/components/StyledTab.tsx (hunk 5 above)
Comment (OlgaMilczek) [reply src/detail-page/components/StyledTab.tsx]: Ok. :)

File: src/detail-page/components/StyledTab.tsx (hunk 7 above)
Comment (OlgaMilczek) [reply src/detail-page/components/StyledTab.tsx]: Done! :)

File: src/detail-page/components/TabsContainer.tsx (hunk 8 above)
Comment (OlgaMilczek) [reply src/detail-page/components/TabsContainer.tsx]: Yes! You are right. Idon't know why I doesn't name it like from the beginning.

File: src/detail-page/components/TabPanel.tsx (hunk 9 above)
Comment (OlgaMilczek) [reply src/detail-page/components/TabPanel.tsx]: Done!

File: src/detail-page/components/TabsContainer.tsx (hunk 10 above)
Comment (OlgaMilczek) [reply src/detail-page/components/TabsContainer.tsx]: Ok. So as I understand I don't need to set any types here? Because TypeScript will do it by itself?

File: src/detail-page/components/TabsContainer.tsx (hunk 11 above)
Comment (OlgaMilczek) [reply src/detail-page/components/TabsContainer.tsx]: Done!

File: src/detail-page/components/StyledTab.tsx (hunk 12 above)
Comment (OlgaMilczek) [reply src/detail-page/components/StyledTab.tsx]: Yes! I is good idea. :)

File: src/detail-page/components/TabsContainer.tsx (hunk 10 above)
Comment (vilanelle) [reply src/detail-page/components/TabsContainer.tsx]: Exactly! We don't need to provide types for params of the callback function passed to `map`, TypeScript will know what types `category` and `id` because of the types we set for `categories` :)

File: src/detail-page/components/TabCategories.tsx (hunk 1 above)
Comment (OlgaMilczek) [reply src/detail-page/components/TabCategories.tsx]: Ok! So for now I move `categories` to props and remove `import { categories } from './tab-components/TabCategories';` in Tabs component. Then I the future we will add it to App state and pass it with the props. 
I move also file `TabCategories` to data folder.

Comment (OlgaMilczek) [reply src/detail-page/components/TabCategories.tsx:19]: I move also file `TabCategories to data folder.

File: src/detail-page/components/TabsContainer.tsx (hunk 13 above)
Comment (OlgaMilczek) [reply src/detail-page/components/TabsContainer.tsx]: Ok! Done :)

File: src/detail-page/components/TabsContainer.tsx (hunk 6 above)
Comment (OlgaMilczek) [reply src/detail-page/components/TabsContainer.tsx]: I change tabContent for routing.

File: .eslintrc.json [hunk 14]
```diff
@@ -66,7 +66,8 @@
     "react/jsx-indent-props": "off",
     "indent": [
       "error",
-      2
+      2,
+      { "SwitchCase": 1 }
```
vilanelle, line 71: :+1:

src/App.tsx
```diff
@@ -21,7 +21,7 @@ export const App: React.FC<{}> = () => (
               <HomePage />
             </Route>
             {ITProfessions.map((r) => (
-              <Route exact path={`/${r}`} key={r}>
+              <Route path={`/${r}`} key={r}>
```
vilanelle: I'd be careful with removing the `exact` param here! :) The routes for professions will be different so we're safe for now, but we may (and probably should) decide to have a common prefix for profession routes (something like `/detail/${profession}`. In this case, we need `exact` to make sure we get to the right routes. More about it here: https://stackoverflow.com/a/49162423.

src/home-page/HomePage.tsx
```diff
@@ -5,12 +5,14 @@ import { Header } from './components/Header';
...
-  <MainContainer>
-    <Header />
-    <Description />
-    <GridContainer>
-      {/* Fill with actual content */}
-    </GridContainer>
-  </MainContainer>
-);
+export const HomePage: React.FC<{}> = () => {
+  return (
```
vilanelle, line 16: If we update the code to master, we will no longer need the explicit return - we have change lint rules.

src/detail-page/components/Tabs.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { useState } from 'react';
```
vilanelle: If we rename our `/tabs-components` folder to simply `/tabs` we can move also this component to this folder :smile_cat:

src/detail-page/components/Tabs.tsx
```diff
@@ -0,0 +1,90 @@
...
+      case 'resources':
+        return <ResourcesTab content={resources} />;
+      default:
+        throw new Error('Wrong category name!');
+    }
+  };
+
+  return (
+    <StyledCard>
+      <StyledTabs indicatorColor="primary" value={chosenCategory}>
```
vilanelle: It looks like we've had a few important changes to `master` meanwhile. Can we please update this branch (it should also allow us to display these new shiny features you added :) )?

src/App.tsx
OlgaMilczek: To be honest I don't know how to make working tabs with Routing without removing exact param here.

klaraborowska: As it was just my silly idea to add routing to those tabs and I didn't know that it can be so problematic in React, I think we should create a new task to work on that issue in the next sprint :)  For now, let's just leave it as it is.

klaraborowska (approved):
Probaby we will require some changes here, when we agree on how/where our content will be stored, but I think it's good for now. Regarding routing, as we discussed, we can create a separate task.

vilanelle (approved):

PR #7, opened 2021-02-12
Drm 7 - github workflow

joannajerz:


.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,23 @@
+name: gh-pages
```
vilanelle, line 1: Wow, I see Github is helping us find any failed workflow check :heart_eyes: 
image
However, the 2 warnings on this PR should disappear after updating to `master` where we have changed lint rules to more sensible :)

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,23 @@
...
+on: [push, pull_request]
```
vilanelle, line 2: Our workflow will now trigger on `push` and `pull_request` which is what we want, but we could configure it more precisely :) Let's add specific branches that we want to target:
```
on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - master
 ```
That will allow us to skip workflow runs every time we push anything to any branch! :zap: We can also make the deploy conditional specifically on `master`, to prevent a deploy if we merge a PR to a different branch:
```
name: Deploy
if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```

File: .github/workflows/gh-pages.yml [hunk 3]
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  build-test:
```
vilanelle, line 5: I think we are being too modest here - this workflow does more than build and test ;) Maybe we can call it `build-release`?

src/App.tsx
```diff
@@ -11,23 +11,27 @@ import { HomePage } from './home-page/HomePage';
...
-            </Route>
-            {ITProfessions.map((r) => (
-              <Route exact path={`/${r}`} key={r}>
-                <p>Works {r}</p>
+export const App: React.FC<{}> = () => {
+  return (
+    <StylesProvider injectFirst>
+      <MuiThemeProvider theme={theme}>
+        <ThemeProvider theme={theme}>
+          <Router basename={window.location.pathname || ''}>
```
vilanelle, line 19: Hm, I'm not sure if we should use `window.location` here - this means we don't have control over our base path. We can use our Github repository path (`tech-transition`) instead (you can have a look at the changes that worked for me here: https://github.com/vilanelle/tech-transition/commit/00a80db89195dbde3aba7aae21264b67cd49cf5b).

vilanelle, line 19: It seems that React Router is not working great with Github pages :/ When we set our basename (either `window.location.pathname` or `/tech-transition`) in `App.tsx`, home page displays fine, but when I try to go to one of our routes it fails: 

image

I tried to find a fix today, but no luck so far. Potential solutions are to set  `PUBLIC_URL` for production environment in `.env` or switch to `HashRouter`. For now, I'd merge it anyway to start benefitting from automated `lint` and `test` checks :clap: And add a separate task to fix routing at some point.

vilanelle (changes requested):

src/App.tsx
OlgaMilczek: There is one problem with card links in master, so that can be also a reason why it doesn't work. Card links us to `public_url/frontend`, but route is for `public_url/details/frontend`

.github/workflows/gh-pages.yml
joannajerz: I think it is done in Workflow routing fix

.github/workflows/gh-pages.yml
joannajerz: It is also done in Workflow routing fix

joannajerz:
Changes in Workflow routing fix

PR #10, opened 2021-02-20
App routes sync

vilanelle:
After a huge amount of work we've done in the past few weeks 💪 we needed to take a step back and to take a look at our app in wholistic manner. As a consequence, I'm proposing a few changes that will help us to align all the bits we prepared in the best manner 🧵


src/App.tsx
```diff
@@ -12,18 +12,13 @@ export const App: React.FC<{}> = () => (
   <StylesProvider injectFirst>
     <MuiThemeProvider theme={theme}>
       <ThemeProvider theme={theme}>
-        <Router>
-          <Switch>
-            <Route exact path="/">
-              <HomePage />
-            </Route>
-            {ITProfessions.map(profession => (
```
vilanelle, line 20: We have no need for a separate `Route` per each profession - we're actually using the same component for all professions (`DetailPage`) so a single `Route` is just fine! :) Also, given that our professions have unique routes, we can use path params to get the information about the current profession.

src/detail-page/DetailPage.tsx
```diff
@@ -22,13 +19,22 @@ export const TabContainer = styled('div')`
...
-}) => (
-  <MainContainer>
-    <div>{profession.title}</div>
-    <TabContainer>
-      <Tabs categories={profession.categories} learn="What to learn" jobs="jobs" resources="resources" />
-    </TabContainer>
-  </MainContainer>
-);
+export const DetailPage: React.FC<{}> = () => {
+  const { professionId } = useParams<{professionId: string}>();
```
vilanelle, line 23: As mentioned above - we don't really need to pass the prop, we can take advantage of information in our params :)

src/detail-page/components/tabs/Tab.tsx
```diff
@@ -11,23 +11,27 @@ const StyledTab = styled(MuiTab)`
...
 
-interface TabProps {
+type TabProps = {
   text: string;
   chosenTab: number;
   index: number;
-  onClick: (index: number) => void;
-}
+  handleClick: (index: number) => void;
+} & MuiTabProps;
```
vilanelle, line 19: We are making our custom `Tab` component, but under the hood it relies on Mui Tab. This is why we want to **preserve** Material-defined props (MuiTabProps) in addition to our custom ones (text, chosenTab, etc). To do this, we changed the type and we use the `...rest` operator to capture any props that we may pass to Tab in Tabs.tsx.

This gives us the following benefit: if in the future we want to modify the tab, we can simply use [Material API](https://material-ui.com/api/tab/). That is, we can pass a Material-defined prop (for example: `disableRipple` or `icon`) to _our_ Tab in Tabs.tsx and the magic of `...rest` will pass it to Material component (StyledTab).

src/detail-page/components/tabs/Tab.tsx
```diff
@@ -11,23 +11,27 @@ const StyledTab = styled(MuiTab)`
...
 
-export const Tab: React.FC<TabProps> = ({
-  text,
-  chosenTab,
-  index,
-  onClick,
-}) => (
-  /* Adding className active to chosen card for creating bold text */
```
vilanelle, line 27: The code is very clean and clear, so I think we can safely remove this comment.

src/detail-page/components/tabs/Tabs.tsx
```diff
@@ -31,60 +31,66 @@ Because it isn't determined how they will look like it's type is unknown.
...
-                text={category.text}
-                chosenTab={chosenCategory}
-                index={index}
-                onClick={setChosenCategory}
-              />
-            </Link>
-          );
-        })}
-      </StyledTabs>
-      {categories.map((category, index) => {
```
vilanelle, line 78: Also here it's enough if we have a single Route (that is we don't have to `map`) - after all, we only want to render one TabPanel at a time! :)

src/detail-page/components/tabs/Tabs.tsx
```diff
@@ -31,60 +31,66 @@ Because it isn't determined how they will look like it's type is unknown.
...
+              chosenTab={activeTab}
+              index={index}
+              handleClick={(tabIndex: number) => handleTabClicked(category, tabIndex)}
+            />
+          ))}
+        </StyledTabs>
+      </StyledCard>
+      <Switch>
+        <Redirect exact from="/details/:professionId" to={`${url}/learn`} />
+        <Route exact path={`${path}/:categoryId`}>
```
vilanelle: There is a difference between `path` and `url` that we get from `useRouteMatch`:
- `url` captures the url at a time, **an exact string**. In this case e.g. `/details/backend`; we should use it when we want to go to a specific url (so e.g. in `Link` or in `history.push()`)
- `path` captures our Route definition, so in this case `/details/:professionId` ◀️ it preserves our **params**. We should use it in `Route`.

src/App.tsx
```diff
@@ -12,18 +12,13 @@ export const App: React.FC<{}> = () => (
...
-              </Route>
-            ))}
-          </Switch>
-        </Router>
+        <Switch>
+          <Route exact path="/">
+            <HomePage />
+          </Route>
+          <Redirect exact path="/details" to={`/details/${ITProfessions[0].route}`} />
+          <Route path="/details/:professionId" component={DetailPage} />
```
vilanelle: As Olga rightly noticed, we can't have `exact` here! :) Why?

We have a `details/:professionId` route which renders `DetailsPage`. `DetailsPage`, in turn, has a child component (`TabPanel`) that will render only if url matches nested route: `details/:professionId/categoryId`, also specified in `DetailsPage`. 

But! If we are on nested route `details/:professionId/categoryId` that means that `DetailsPage` route will not be matched if we use `exact` (`details/:professionId/categoryId` !== `details/:professionId`). So `DetailsPage` never renders which means that our nested route definition doesn't render either. Effectively, our nested route **doesn't exist** 🤷

src/detail-page/components/tabs/Tabs.tsx
```diff
@@ -31,60 +31,66 @@ Because it isn't determined how they will look like it's type is unknown.
...
+    setActiveTab(index);
   };
 
   return (
-    <StyledCard>
-      <StyledTabs indicatorColor="primary" value={chosenCategory}>
-        {categories.map((category, index) => {
-          const urlPath = index === 0 ? match.url : `${match.url}/${category.name}`;
-          return (
-            <Link to={`${urlPath}`} key={category.name}>
```
vilanelle, line 67: Looks like Material `Tabs` is unhappy if it gets children other than `Tab` components:

image

But navigating programatically by using `history.push()` works fine! :)

src/App.tsx
```diff
@@ -12,18 +12,13 @@ export const App: React.FC<{}> = () => (
...
-                <DetailPage profession={profession} />
-              </Route>
-            ))}
-          </Switch>
-        </Router>
+        <Switch>
+          <Route exact path="/">
+            <HomePage />
+          </Route>
+          <Redirect exact path="/details" to={`/details/${ITProfessions[0].route}`} />
```
vilanelle: This redirect will navigate to first profession url if the user tries to enter `details` route without any profession id :)

src/detail-page/components/tabs/Tabs.tsx
```diff
@@ -31,60 +31,66 @@ Because it isn't determined how they will look like it's type is unknown.
...
+              text={category.text}
+              chosenTab={activeTab}
+              index={index}
+              handleClick={(tabIndex: number) => handleTabClicked(category, tabIndex)}
+            />
+          ))}
+        </StyledTabs>
+      </StyledCard>
+      <Switch>
+        <Redirect exact from="/details/:professionId" to={`${url}/learn`} />
```
vilanelle: This is to make sure that users will have a proper url (first tab) if they enter DetailsPage url without any tab param.

src/data/ITProfessions.ts
```diff
@@ -8,11 +8,12 @@ import uxIcon from '../icons/ux_ui.svg';
 export type ProfessionKey = 'frontend' | 'backend' | 'uxui' | 'qa' | 'devops' | 'projectmanager';
 
 export type Category = {
```
OlgaMilczek, line 10: I see we are using interface and type for objects interchangeably. I think we should unify it some how and use inteface for objects and type for variables (I know it was me who use type here :D )

src/App.tsx
```diff
@@ -17,11 +17,8 @@ export const App: React.FC<{}> = () => (
             <Route exact path="/">
               <HomePage />
             </Route>
-            {ITProfessions.map(profession => (
-              <Route path={`/details/${profession.route}`} key={profession.id}>
-                <DetailPage profession={profession} />
-              </Route>
-            ))}
+            <Redirect exact path="/details" to={`/details/${ITProfessions[0].route}`} />
+            <Route path="/details/:professionId" component={DetailPage} />
```
OlgaMilczek, line 23: Maybe we can add some comment here? Because I think it isn't obvious and maybe in the future someone else will work on it and will need to spend some time to understand this code.

src/detail-page/DetailPage.tsx
OlgaMilczek: That is great! :)

src/data/ITProfessions.ts
OlgaMilczek: Now I think we should talk about it with other girls and choose a common strategy :)

src/detail-page/components/tabs/Tabs.tsx
```diff
@@ -1,14 +1,14 @@
 import React, { useState } from 'react';
-import { Route, Link, useRouteMatch } from 'react-router-dom';
+import { Redirect, Route, Switch, useHistory, useRouteMatch } from 'react-router-dom';
 import { Card, Tabs as MuiTabs } from '@material-ui/core';
 import styled from 'styled-components';
 
 import { TabPanel } from './TabPanel';
 import { Tab } from './Tab';
-import { ResourcesTab } from '../ResourcesTab';
+import { Category, Profession } from '../../../data/ITProfessions';
```
OlgaMilczek, line 8: Maybe we can separate data import form React Components import? I mean to have one group of import for React Components and separate one for data (event when it is only one import).

src/detail-page/components/tabs/Tabs.tsx
```diff
@@ -31,60 +31,66 @@ Because it isn't determined how they will look like it's type is unknown.
...
 
-export const Tabs: React.FC<TabProps> = ({
-  categories,
-  learn,
-  jobs,
-  resources,
-}) => {
-  const [chosenCategory, setChosenCategory] = useState(0);
-  const match = useRouteMatch();
+const tabList = {
```
OlgaMilczek, line 34: Maybe we can add comment here that this is temporary and should be replace with actual content of Tabs?

src/detail-page/components/tabs/Tabs.tsx
```diff
@@ -31,60 +31,66 @@ Because it isn't determined how they will look like it's type is unknown.
...
+      <Switch>
+        <Redirect exact from="/details/:professionId" to={`${url}/learn`} />
+        <Route exact path={`${path}/:categoryId`}>
+          <TabPanel
+            getTabContent={getTabContent}
+            syncTabWithPath={syncTabWithPath}
+          />
+        </Route>
+      </Switch>
+    </>
```
OlgaMilczek, line 91: I have a question here. 

Why our TabPanel is outside StyledCard? Because now our Tabs have this rounded border and some smooth shadow on the bottom. When we will add some content in the next tasks to the TabPanels it will look a little bit weird. I think Tabs and TabPanel content sholud be wrapped in one Card content or some div. But maybe I am missing somoething here.

src/data/ITProfessions.ts
vilanelle: Good point! And a great idea to find a consistent solution together :muscle: 
So I'll probably do some clean-up here, but we'll will talk about it and make the decisions at the next meeting :)

src/App.tsx
vilanelle: Hm, that's a good point. However, to avoid a comment I thought of another solution: I put the `to=` path in a constant variable and named it properly (`DefaultDetailsPagePath`). Hope that's better now!

src/detail-page/components/tabs/Tabs.tsx
vilanelle: Sure!

src/detail-page/components/tabs/Tabs.tsx
vilanelle: Done! :)

src/detail-page/components/tabs/Tabs.tsx
vilanelle: Ah, that's my omission. Fixed (and renamed `StyledCard` to `TabsContainer` to make it a bit more meaningful).

joannajerz (approved):

OlgaMilczek (approved):

PR #11, opened 2021-02-21
Workflow routing fix

vilanelle:


.eslintrc.json
```diff
@@ -54,6 +54,7 @@
     "@typescript-eslint/prefer-interface": "off",
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-non-null-assertion": "off",
+    "@typescript-eslint/no-unused-vars": "off",
```
vilanelle, line 57: Our linter complains about unused vars when we import something we use only for typing (e.g. `MuiTabs` in `Tab.tsx`). I disabled the rule temporarily before we can fix this in a more proper manner.

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,31 @@
+name: gh-pages
+on:
```
vilanelle, line 2: The `on` section allows us to trigger workflow depending on developer action and on specific branches. Currently, our setup will trigger the workflow when someone pushes (merges) changes to `master` or makes a PR to `master`.

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Set-up Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: "10.x"
+      - run: npm install
+      - run: npm run build
+      - run: npm run test
+      - run: npm run lint
+      - name: Deploy
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```
vilanelle, line 27: In the `on` section, we trigger workflow on a PR to master to make sure that PRs contain code that passes tests and lint. However, we don't want such code to deploy to production (in our case, to Github Pages)! It may not be ready and it hasn't passed the review. This is why we add a condition for the `Deploy` step - now the deploys step will run only when we push to `master` :)

package.json
```diff
@@ -31,15 +31,18 @@
...
     "eslint-plugin-react": "^7.22.0",
-    "eslint-plugin-react-hooks": "^4.2.0"
+    "eslint-plugin-react-hooks": "^4.2.0",
+    "dotenv-cli": "^4.0.0",
+    "gh-pages": "^3.1.0"
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "dotenv -e .env.production react-scripts build",
```
vilanelle, line 40: Turns out we need a separate utility (`dotenv`) to set different `PUBLIC_URL` per environment. We set our config in a separate file (`.env.production`) and pass this file to `dotenv` in build step.

src/App.tsx
```diff
@@ -12,7 +12,7 @@ export const App: React.FC<{}> = () => (
   <StylesProvider injectFirst>
     <MuiThemeProvider theme={theme}>
       <ThemeProvider theme={theme}>
-        <Router>
+        <Router basename={`${process.env.PUBLIC_URL}`}>
```
vilanelle, line 15: We need to add Github repo name (`tech-transition`) to our paths to take into account how Github constructs pages urls. Otherwise, the routing will not work after deployment.

joannajerz (approved):

PR #13, opened 2021-02-22
Drm 13 back button

joannajerz:


ireshka (commented):
After merging big changes from previous PR we should synchronize branch with actual master. :) Back button component looks good but sth is wrong with routing.

src/detail-page/BackButton.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import styled from 'styled-components';
+import Typography from '@material-ui/core/Typography';
+import ArrowBackIosIcon from '@material-ui/icons/ArrowBackIos';
+import { Button } from '@material-ui/core';
+import { useHistory } from 'react-router-dom';
+
+const BackButtonStyle = styled(Button)`
```
vilanelle, line 8: Maybe `BackButton` is enough (shorter name)? :)

.gitignore
```diff
@@ -25,4 +25,4 @@ yarn-error.log*
 .eslintcache
```
vilanelle: We had a lot of changes in our code recently, so it would be good if we updated this branch to `master`.
We can use `merge` or `cherry-pick` (`rebase` is not the best option here, because of the number of commits that we need to incorporate in the current branch).

src/detail-page/BackButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+  margin: 1em;
```
vilanelle: We could use our theme here to keep margins between page elements nice and consistent across the app :)
```
const BackButtonStyle = styled(Button)`
  margin: ${props => props.theme.spacing(2)}px;  
`;
```

vilanelle (changes requested):

src/detail-page/BackButton.tsx
joannajerz: I have another const (Backbutton) lower so i have a name conflict when i change.

src/detail-page/BackButton.tsx
klaraborowska: I think it would look nice, if the button was aligned to the left with tabs container, so maybe this margin could be on the top and bottom only?

joannajerz: I think it's done

src/detail-page/BackButton.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import styled from 'styled-components';
+import Typography from '@material-ui/core/Typography';
+import ArrowBackIosIcon from '@material-ui/icons/ArrowBackIos';
+import { Button } from '@material-ui/core';
+import { useHistory } from 'react-router-dom';
+
+const BackButtonStyle = styled(Button)`
+margin: ${props => props.theme.spacing(2)}px 0px;
```
klaraborowska: looks good now :) the only thing - we don't need to add `px` when value is 0

klaraborowska (approved):

src/detail-page/BackButton.tsx
joannajerz: done

vilanelle (approved):

PR #14, opened 2021-02-26
DRM-16 - Reuse Card Component & DRM-41 - Extra wrapper in main container

ireshka:
PR combines two tasks - adding card component to Details view and unifying content container.


src/detail-page/DetailPage.tsx
```diff
@@ -6,15 +6,14 @@ import { MainContainer } from '../layout/components/MainContainer';
 import { Tabs } from './components/tabs/Tabs';
 
 import { ITProfessions } from '../data/ITProfessions';
+import { ProfessionCard } from '../home-page/components/ProfessionCard';
```
ireshka: We use ProfessionCard in two views now - HomePage and DetailPage. Should we move ProfessionCard outside of HomePage or we keep it there?

tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react"
+    "jsx": "react-jsx"
```
ireshka: After starting project by `npm run start` it changed it automatically from `react` to `react-jsx`

src/layout/components/MainContainer.tsx
```diff
@@ -1,7 +1,17 @@
...
-    min-height: 100vh;
+const BackgroundContainer = styled('div')`
+  background: url('${process.env.PUBLIC_URL}/img/background-img.png') center top / cover no-repeat;
+  min-height: 100vh;
 `;
+
+export const MainContainer: React.FC<{}> = ({ children }) => (
+  <BackgroundContainer>
+    <Container maxWidth="md">
+      { children! }
```
ireshka, line 23: It solves issue described here: https://github.com/mui-org/material-ui/issues/21711

src/home-page/components/ProfessionCard.tsx
```diff
@@ -21,16 +21,16 @@ const CardHeaderContainer = styled(CardHeader)`
 const AvatarContainer = styled(Avatar)<{ background: ProfessionKey }>`
   margin-right: 1em;
   background-color: ${({ background, theme }) => theme.palette.professions[background].main};
-  padding: .75em;
+  padding: .35em;
```
ireshka: I think smaller padding looks better. What do you think about?

src/detail-page/DetailPage.tsx
```diff
@@ -6,15 +6,14 @@ import { MainContainer } from '../layout/components/MainContainer';
...
 
 export const TabContainer = styled('div')`
-    max-width: 960px;
-    width: 80%;
-    margin: 2em auto 0;
+    width: 100%;
+    margin: 2em 0 0;
```
vilanelle: Maybe we could start using our theme values to keep spacing consistent across the app? Like so:
```
margin: ${props => props.theme.spacing(2)}px 0 0;  
```

src/detail-page/DetailPage.tsx
vilanelle: Excellent point! It looks like we don't have a proper place in the app to put this component so we should create a new one. `/shared` folder is one possible option :)

src/detail-page/DetailPage.tsx
```diff
@@ -26,7 +25,15 @@ export const DetailPage: React.FC<{}> = () => {
     <>
       {profession && (
         <MainContainer>
-          <div>{profession.title}</div>
+          <ProfessionCard
```
vilanelle, line 28: The card here is not clickable, but it has styling suggesting an interactive element (opacity changes on hover, cursor pointer). We should manage it, so that we don't suggest to the user that `DetailsPage` card is interactive. At the same time we need to make sure that we preserve these styles on `HomePage`, where the card is clickable :upside_down_face:

src/home-page/components/ProfessionCard.tsx
```diff
@@ -41,7 +41,7 @@ export const ProfessionCard: React.FC<CardProps> = ({
   clickHandler,
   route,
 }) => (
-  <CategoryCardContainer onClick={() => clickHandler(route)}>
+  <CategoryCardContainer onClick={() => (clickHandler && route) ? clickHandler(route) : null}>
```
vilanelle: This conditional function call is a warning sign that maybe we should take a step back and reconsider if everything is well here. Maybe we could lift the click handler to the parent, that is do not pass it, but keep the logic one level up? Or maybe it's better to duplicate the Component?
Also, could we change the name of the container to `ProfessionCardContainer` (my bad, I missed it when doing the first change)?

tsconfig.json
vilanelle: It looks like VS Code issue, have a look at the solution suggested here: 

https://stackoverflow.com/questions/64974648/problem-with-visual-studio-code-using-react-jsx-as-jsx-value-with-create-react

Hopefully it helps! :)

vilanelle (changes requested):

src/detail-page/DetailPage.tsx
```diff
@@ -26,7 +25,15 @@ export const DetailPage: React.FC<{}> = () => {
     <>
       {profession && (
         <MainContainer>
```
klaraborowska, line 27: We need some space between the wrapper and card, now card displays on the top of the website . Maybe we could add padding-top to the wrapper?

src/detail-page/DetailPage.tsx
ireshka: I moved ProfessionCard to shared directory.

src/detail-page/DetailPage.tsx
ireshka: I tried unify spacing with theme values - it may be a good idea to check it everywhere after we merge all changes.

src/detail-page/DetailPage.tsx
ireshka: I added top padding to the content container.

src/home-page/components/ProfessionCard.tsx
ireshka: I refactored ProfessionCard. Please check new changes. :)

src/detail-page/DetailPage.tsx
```diff
@@ -1,22 +1,23 @@
...
+    margin: ${theme.spacing(4, 0, 0)};
+    padding-bottom: ${theme.spacing(4)}px;
 
-    ${props => props.theme.breakpoints.down('sm')} {
-      width: 90%;
-      margin-top: 1.5em;
+    ${theme.breakpoints.down('sm')} {
+      margin: ${theme.spacing(3, 0, 0)};
     }
+  `}
```
ireshka: I changed spacing here to defaut theme spacing. We may later discuss with all girls if we want to use absolute or relative values in spacing and if we want to change something more - adjust theme.

src/home-page/HomePage.tsx
```diff
@@ -1,35 +1,39 @@
...
             avatarBackground={color}
-            clickHandler={handleClick}
-            route={route}
           />
-        ))}
-      </GridContainer>
-    </MainContainer>
-  );
-};
+        </StyledLink>
```
ireshka, line 35: By using Link element from router we can use card as link - for example open in new tab with Ctrl+Click. icClickable props is essential for right styling clickable/non clickable version of ProfessionCard.

src/shared/components/ProfessionCard.tsx
```diff
@@ -21,28 +20,26 @@ const CardHeaderContainer = styled(CardHeader)`
...
-  route,
+  isClickable,
 }) => (
-  <CategoryCardContainer onClick={() => clickHandler(route)}>
+  <ProfessionCardContainer $isClickable={isClickable !== undefined}>
     <CardHeaderContainer title={title} subheader={description} />
     <AvatarContainer src={avatarSrc} background={avatarBackground} />
-  </CategoryCardContainer>
+  </ProfessionCardContainer>
 );
```
ireshka: I use here - 'transient props' - https://github.com/styled-components/styled-components/releases/tag/v5.1.0
It solves problem with passing props that are needed only for styling purpose - not for whole component.

src/detail-page/DetailPage.tsx
```diff
@@ -1,22 +1,23 @@
...
 
 import { ITProfessions } from '../data/ITProfessions';
+import { ProfessionCard } from '../shared/components/ProfessionCard';
 
 export const TabContainer = styled('div')`
-    max-width: 960px;
-    width: 80%;
-    margin: 2em auto 0;
-    padding-bottom: 2em;
+  ${({ theme }) => css`
```
vilanelle: Do we need the `css`? We can get `theme` from function props, like we did previously?

src/shared/components/ProfessionCard.tsx
```diff
@@ -21,28 +20,26 @@ const CardHeaderContainer = styled(CardHeader)`
...
   title,
   description,
   avatarSrc,
   avatarBackground,
-  clickHandler,
-  route,
+  isClickable,
 }) => (
-  <CategoryCardContainer onClick={() => clickHandler(route)}>
+  <ProfessionCardContainer $isClickable={isClickable !== undefined}>
```
vilanelle: If we make `isClickable` a non-optional prop, we can remove this `!==` :)

src/shared/components/ProfessionCard.tsx
```diff
@@ -3,14 +3,13 @@ import { Card, CardHeader, Avatar } from '@material-ui/core';
 import styled from 'styled-components';
 import { ProfessionKey } from '../../data/ITProfessions';
 
-const CategoryCardContainer = styled(Card)`
+const ProfessionCardContainer = styled(Card)<{ $isClickable: boolean }>`
```
vilanelle: I know it's not part of your PR, but maybe we could rename the `Container` suffixes in this file to `Styled` prefixes? For example, `ProfessionCard` rather than `ProfessionCardContainer`. 
It may seem very picky on my part, but I have a good reason to suggest it: by convention `container` is a React component that has logic, state etc. as opposed to the so called "dumb" component that has no logic. When you are used to this, seeing 4 container components in one file is a bit confusing :)

vilanelle (approved):

src/shared/components/ProfessionCard.tsx
ireshka: When we set isClickable as required prop we have to remember about it in every component. With optional isClickable we can add it only with component with 'clickable style'. That was an idea, why isClickable is optional :)

klaraborowska: You can use default props in here `export const ProfessionCard: React.FC<ProfessionCardProps> = ({
  title,
  description,
  avatarSrc,
  avatarBackground,
  isClickable = false,
})`
This way we can get rid of `undefined` and we don't have to add `isClickable` each time.

klaraborowska (approved):

src/shared/components/ProfessionCard.tsx
ireshka: @klaraborowska I used your advice but when I set `isClickable` as required in type `ProfessionCard` typescript still complains about missing props if I don't provide it explicitly (even if I set default props).
I join your hints - and `isClickable` is an optional prop with default value `false`.

klaraborowska: Maybe I didn't write it clear enough, but you did exactly what I meant ;-)

src/layout/components/MainContainer.tsx
PiotrGrobelak: thx a lot :)

PR #15, opened 2021-02-28
Drm 57 404 page

joannajerz:


src/detail-page/DetailPage.tsx
```diff
@@ -20,14 +21,15 @@ export const TabContainer = styled('div')`
 `;
 
 export const DetailPage: React.FC<{}> = () => {
-  const { professionId } = useParams<{professionId: string}>();
+  const { professionId } = useParams<{ professionId: string }>();
   const profession = ITProfessions.find(p => p.id === professionId);
```
klaraborowska, line 25: I think it would be good to redirect to the error page, when there is no profession.
The same thing we could do with tabs. For now we throw new error, when someone for example type a tabname, which we don't have. We could change it for redirection to error page.

src/home-page/NoMatch404.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
```
klaraborowska: I really like the look of 404 page! :)

src/home-page/NoMatch404.tsx
```diff
@@ -0,0 +1,41 @@
...
+color: white;
+`;
+
+export const NoMatch404: React.FC<{}> = () => {
+  const history = useHistory();
+
+  return (
+    <MainContainer>
+      <NoMatchTypography>
+        <HeaderTypography variant="h1">404</HeaderTypography>
```
klaraborowska: Sorry, I didn't notice it before. I think it shuldn't be h1. H1 should be reserved for the main information on our page, like the main heading.

src/detail-page/DetailPage.tsx
joannajerz: done

src/home-page/NoMatch404.tsx
joannajerz: done;)

src/home-page/NoMatch404.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { useHistory } from 'react-router-dom';
+import Typography from '@material-ui/core/Typography';
+import styled from 'styled-components';
+import { Button } from '@material-ui/core';
+import { MainContainer } from '../layout/components/MainContainer';
+
+const NoMatchTypography = styled('div')`
```
OlgaMilczek: I think this name here "NoMatchTypography" can be a bit confusing, because this component is more some kind of container, then Typography. In my opinion name Typography points more on some kind of pure text element. Maybe we can rename it to be more suitable and descriptive. Maybe 404Card or NoMatchCard, what do you think?

src/home-page/NoMatch404.tsx
```diff
@@ -0,0 +1,41 @@
...
+const HeaderTypography = styled(Typography)`
+  font-size:10em;
+`;
+const SubtitleTypography = styled(Typography)`
+  font-size:2em;
+  margin: ${props => props.theme.spacing(2)}px;  
+`;
+
+const StyledButton = styled(Button)`
+color: white;
```
OlgaMilczek: Maybe we can use here our them variables as well (I mean add white to our them and use it here instead hard code white here). It will be easier in the future to change colors in one place then looking for them in different components. :)

src/detail-page/components/tabs/Tabs.tsx
```diff
@@ -46,7 +47,7 @@ const getTabContent = (categoryName: String) => {
     case 'resources':
       return <ResourcesTab content={tabList.resources} />;
     default:
-      throw new Error('Wrong category name!');
+      return <Redirect to="/404" />;
```
OlgaMilczek, line 49: :+1:

src/App.tsx
```diff
@@ -7,6 +7,7 @@ import { HomePage } from './home-page/HomePage';
 import { theme } from './layout/theme';
 import { DetailPage } from './detail-page/DetailPage';
 import { ITProfessions } from './data/ITProfessions';
+import { NoMatch404 } from './home-page/NoMatch404';
```
OlgaMilczek: We have folder "src/layout/components" for components which will be used in different places in the page (MainContainer is there). Maybe it will be better place for 404 page component because it is not specific for home page only.

vilanelle: That's a good point! Alternatively we could set up a separate folder for error pages. But for now we will only have 404, so it seems an overkill.

src/home-page/NoMatch404.tsx
vilanelle: I'd recommend simply a `CardWrapper` :)

src/home-page/NoMatch404.tsx
```diff
@@ -0,0 +1,41 @@
...
+  margin: auto;
+  min-height: 70vh;
+`;
+
+const HeaderTypography = styled(Typography)`
+  font-size:10em;
+`;
+const SubtitleTypography = styled(Typography)`
+  font-size:2em;
+  margin: ${props => props.theme.spacing(2)}px;
```
vilanelle: It's great to see that we're using uniform spacing from our theme! ♥

src/home-page/NoMatch404.tsx
```diff
@@ -0,0 +1,41 @@
...
+  display:flex;
+  justify-content: center;
+  align-items: center;
+  flex-direction:column;
+  margin: auto;
+  min-height: 70vh;
+`;
+
+const HeaderTypography = styled(Typography)`
+  font-size:10em;
```
vilanelle: I know it seems like a very small thing, but could we add a space between css property and value (also for SubtitleTypography)?

src/home-page/NoMatch404.tsx
```diff
@@ -0,0 +1,41 @@
...
+const NoMatchTypography = styled('div')`
+  display:flex;
+  justify-content: center;
+  align-items: center;
+  flex-direction:column;
+  margin: auto;
+  min-height: 70vh;
+`;
+
+const HeaderTypography = styled(Typography)`
```
vilanelle: Mui `Typography` is a way to keep text (fonts, etc.) consistent across our app. Consequently, we should not override font size for individual `Typography` components, but rather choose font sizes for HTML elements (`h1`, `p` etc.) in our Theme and then apply `Typography` variants. 
So the best option here seems to set the desired font size in Theme, say enlarging `h1` and adding a value for `subtitle1`:
```export const theme = createMuiTheme({
  typography: {
    fontFamily: 'Segoe UI, sans-serif',
    h1: {
      fontSize: 40, // <- change to bigger font
      [breakpoints.down('sm')]: {
        fontSize: 30,
      },
      [breakpoints.down('xs')]: {
        fontSize: 24,
      },
    },
    subtitle1: { 
      fontSize: 20, // add a fitting font size for subtitle
    },
    body1: {
      [breakpoints.down('sm')]: {
        fontSize: 14,
      },
      [breakpoints.down('xs')]: {
        fontSize: 13,
      },
    },
```

That way we make sure all HTML elements will have a consistent size across the app with minimum effort! 💪

src/home-page/NoMatch404.tsx
```diff
@@ -0,0 +1,41 @@
...
+`;
+
+export const NoMatch404: React.FC<{}> = () => {
+  const history = useHistory();
+
+  return (
+    <MainContainer>
+      <NoMatchTypography>
+        <HeaderTypography variant="h4">404</HeaderTypography>
+        <SubtitleTypography variant="subtitle1">It looks like you&apos;re lost.</SubtitleTypography>
```
vilanelle: Let's make use of Material Ui building blocks: instead of adding `SubtitleTypography` margin via `styled`, we can grab Mui `Box` component that offers a great API for spacing and sizing. Assuming we also move font size management to Theme, we end up with:
```
<Typography variant="h1">404</Typography>
<Box m={2}>
  <Typography variant="subtitle1">It looks like you&apos;re lost.</Typography>
</Box>
```
☝ we can remove `SubtitleTypography` and `HeaderTypography` and as a bonus we achieve a nice separation of concerns (`Box` is responsible for spacing, Typography for fonts) 🙃

src/home-page/NoMatch404.tsx
joannajerz: Done

src/App.tsx
joannajerz: done

src/home-page/NoMatch404.tsx
joannajerz: done

src/home-page/NoMatch404.tsx
joannajerz: done

src/home-page/NoMatch404.tsx
joannajerz: done

src/home-page/NoMatch404.tsx
joannajerz: done

OlgaMilczek (approved):

vilanelle (approved):

PR #16, opened 2021-03-11
DRM 60 - Add chart js library to the project

Stajgar:
https://dareit.atlassian.net/browse/DRM-60

After: 
- Added Chart.js and React wrapper libraries to the project
- Added demo chart to detail page


klaraborowska (approved):

src/home-page/components/DemoChart.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { Bar } from 'react-chartjs-2';
+
+export const DemoChart: React.FC<{}> = () => (
+  <div>
+    <Bar
+      data={{
```
klaraborowska: I think it could be better to read, if we declare the data and options on the top and use it here for example like this `data={data}`

src/detail-page/DetailPage.tsx
```diff
@@ -40,6 +41,7 @@ export const DetailPage: React.FC<{}> = () => {
               profession={profession}
             />
           </TabContainer>
+          <DemoChart />
```
OlgaMilczek: Maybe we can add some comment here and before import that this is temporary? :)

src/detail-page/DetailPage.tsx
```diff
@@ -7,6 +7,7 @@ import { Tabs } from './components/tabs/Tabs';
 
 import { ITProfessions } from '../data/ITProfessions';
 import { ProfessionCard } from '../shared/components/ProfessionCard';
+import { DemoChart } from '../home-page/components/DemoChart';
```
OlgaMilczek: I think detail-page folder will be more suitable for DemoChart component because it is not use in home-page :)

src/home-page/components/DemoChart.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { Bar } from 'react-chartjs-2';
+
+export const DemoChart: React.FC<{}> = () => (
```
vilanelle: It looks lovely 😍
image

src/detail-page/DetailPage.tsx
vilanelle: I'd even go a step further and remove the use of component (line 44) before merge to `master`. After all, we have a live production site, so we should take care of what we release 😎

src/home-page/components/DemoChart.tsx
Stajgar: Good idea. It's done.

src/detail-page/DetailPage.tsx
Stajgar: Good point, I've must have missed it. It's moved.

src/detail-page/DetailPage.tsx
Stajgar: Done

OlgaMilczek (approved):

PR #17, opened 2021-03-13
Drm 47 find job tab content

joannajerz:


src/data/ITProfessions.ts
```diff
@@ -22,6 +22,13 @@ export interface Profession {
   route: ProfessionKey;
   categories: Category[];
 }
+export interface Job {
+  company: string;
+  title: string;
+  description: string;
+  url: string;
+  id: string;
+}
```
OlgaMilczek: Maybe we can do separate file in data folder for this Interface, because it is not use here and this file is already very long?

src/detail-page/components/JobsTab.tsx
```diff
@@ -1,12 +1,13 @@
 import React from 'react';
+import FetchJobContent from './TabJobContent/FetchJobContent';
```
OlgaMilczek: In my opinion name FetchJobContent is more suited for a fetching function than for the component that retrieves and displays fetched data. Moreover JobTab is now a redundant component. Maybe we should move content from FetchJobContent to JobTab? Because this is the actual content of this Tab :)

src/detail-page/components/JobsTab.tsx
```diff
@@ -1,12 +1,13 @@
...
 
 interface JobProps {
   content: string;
 }
 
 export const JobsTab: React.FC<JobProps> = ({ content }) => (
```
OlgaMilczek: Content is not needed here :) This was only to show that Tabs are working correctly. Also see line below.

src/detail-page/components/TabJobContent/FetchJobContent.tsx
```diff
@@ -0,0 +1,79 @@
+import { Button, CircularProgress, Input } from '@material-ui/core';
+import styled from 'styled-components';
+import * as React from 'react';
+import { Redirect } from 'react-router-dom';
+import { Job } from '../../../data/ITProfessions';
+import FetchJobElement from './FetchJobElement';
+import hiringPicture from '../../../icons/hiring.svg';
```
OlgaMilczek: Maybe we should organize import somehow. Frameworks import in the beginning (about React), library second (Material UI) and then data and our components and images in the end.

src/detail-page/components/TabJobContent/FetchJobContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+          }
+          )),
+        );
+        setFetching(false);
+      }).catch(() => {
+        <Redirect to="/404" />;
+      });
+    console.log(jobPosts);
+  }, [role, city, jobPosts]);
+
```
OlgaMilczek: What do you think about writing separate async function and passing it here in React.useCallback? In my opinion js await function is more readable :)

src/detail-page/components/TabJobContent/FetchJobContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+width:50%;
+`;
+
+const FetchJobContent: React.FC = () => {
+  const [jobPosts, setJobPosts] = React.useState([]);
+  const [role, setRole] = React.useState('');
+  const [city, setCity] = React.useState('');
+  const [fetching, setFetching] = React.useState(false);
+  const handleChangeRole = React.useCallback((e) => setRole(e.target.value), []);
+  const handleChangeCity = React.useCallback((e) => setCity(e.target.value), []);
```
OlgaMilczek: I never use useCallback hook and I don't understand how it works. Can you explain how it works and why you use it here? It doesn't mean I suggest it is something wrong with this code. I just want to understand this solution :)

src/detail-page/components/TabJobContent/FetchJobContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  const handleChangeJobs = React.useCallback(() => {
+    setFetching(true);
+    fetch(`https://cors.bridged.cc/https://jobs.github.com/positions.json?title=${role}&location=${city}`)
+      .then((response) => response.json())
```
OlgaMilczek: I think it is more readable when we don't use parentheses as we only have one argument. :) Or this is again something with our Linter?

src/detail-page/components/TabJobContent/FetchJobContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+      <HiringImage data={hiringPicture} />
+      {fetching ? <SearchSpinner><CircularProgress /></SearchSpinner> : null}
+      <ul>
+        {jobPosts.map((job) => <FetchJobElement job={job} />)}
+      </ul>
+    </>
+  );
+};
+
+export default FetchJobContent;
```
OlgaMilczek: Did you try to fetch some data form this database? I've tried "Engineer" and "USA" and the responses contain HTML which gets escaped - maybe we can do better?

Moreover we can add a jira task for someone to handle the "no results" case.

src/detail-page/components/TabJobContent/FetchJobContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  return (
+    <>
+      <SearchInput type="search" placeholder="Job title" value={role} disableUnderline onChange={handleChangeRole} />
+      <SearchInput placeholder="City" disableUnderline value={city} onChange={handleChangeCity} />
+      <SearchButton color="secondary" variant="contained" onClick={handleChangeJobs}>Search</SearchButton>
+      <HiringImage data={hiringPicture} />
+      {fetching ? <SearchSpinner><CircularProgress /></SearchSpinner> : null}
+      <ul>
+        {jobPosts.map((job) => <FetchJobElement job={job} />)}
```
OlgaMilczek: We could add a "key" prop to avoid the warning:
`Warning: Each child in a list should have a unique "key" prop.`

src/detail-page/components/TabJobContent/FetchJobElement.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+import { Button } from '@material-ui/core';
+import { Job } from '../../../data/ITProfessions';
+
+const FetchJobElement = (props: { job: Job }) => {
```
OlgaMilczek: Similarly to FetchJobContent, this name sounds too function-esque too me.
Would you object to something like JobCard?

src/detail-page/components/TabJobContent/FetchJobContent.tsx
vilanelle: Just a small reminder that if we think we could improve our lint rules, we are free to do so :))

src/detail-page/components/TabJobContent/FetchJobContent.tsx
```diff
@@ -0,0 +1,79 @@
+import { Button, CircularProgress, Input } from '@material-ui/core';
+import styled from 'styled-components';
+import * as React from 'react';
+import { Redirect } from 'react-router-dom';
+import { Job } from '../../../data/ITProfessions';
+import FetchJobElement from './FetchJobElement';
+import hiringPicture from '../../../icons/hiring.svg';
+
+const SearchInput = styled(Input)`
+width: 250px;
```
vilanelle: Let's clean up indentation and spacing for styled components in this file - it'll make our styles much easier to read! :)

src/detail-page/components/TabJobContent/FetchJobContent.tsx
vilanelle: This task was to fetch results only, so I think it's a good idea to add tasks for presentation of results and handling of edge cases such as no results found :)

src/detail-page/components/TabJobContent/FetchJobElement.tsx
vilanelle: There is a good reason why we should not use `fetch` in the component name - this component is very simple (dumb component) and should have no knowledge about anything outside. Specifically, it should have no knowledge that it is being fetched :)
I'd go with `JobListItem` (`Card` could be confusing because we use Material Cards elsewhere and this is a `<li>`)

src/detail-page/components/TabJobContent/FetchJobElement.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
```
vilanelle: It's great that we extracted a reusable, independent part to a separate component! 👏

src/detail-page/components/TabJobContent/FetchJobElement.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { Button } from '@material-ui/core';
+import { Job } from '../../../data/ITProfessions';
+
```
vilanelle: We could add types for component props - then we can destructure props nicely:
```
interface Props {
  job: Job;
}

const FetchJobElement: React.FC<Props> = ({ job }) => { ...}

File: src/detail-page/components/TabJobContent/FetchJobContent.tsx [hunk 14]
```diff
@@ -0,0 +1,79 @@
...
+            title: job.title,
+            description: job.description,
+            url: job.url,
+            key: job.id,
+          }
+          )),
+        );
+        setFetching(false);
+      }).catch(() => {
+        <Redirect to="/404" />;
```
vilanelle: I'm not sure if we want to redirect the user to 404, when any error happens. Our 404 page should be rather a safeguard to catch users entering wrong url.  Also, the `catch` clause can mean many other errors apart from 404.
An error notification (such as a toast) would be a better option here. We can handle it in a separate task / PR.

src/detail-page/components/TabJobContent/FetchJobContent.tsx
```diff
@@ -0,0 +1,79 @@
+import { Button, CircularProgress, Input } from '@material-ui/core';
+import styled from 'styled-components';
+import * as React from 'react';
```
vilanelle: There's no need for the `*` - we can simply write:
```
import React from 'react';
```
We can also add named imports for `useState` and `useCallback`, that will allow us to remove the repeated `React` prefix 🙃
```
import React, { useState, useCallback } from 'react';

File: src/detail-page/components/TabJobContent/FetchJobContent.tsx [hunk 16]
```diff
@@ -0,0 +1,79 @@
...
+
+const FetchJobContent: React.FC = () => {
+  const [jobPosts, setJobPosts] = React.useState([]);
+  const [role, setRole] = React.useState('');
+  const [city, setCity] = React.useState('');
+  const [fetching, setFetching] = React.useState(false);
+  const handleChangeRole = React.useCallback((e) => setRole(e.target.value), []);
+  const handleChangeCity = React.useCallback((e) => setCity(e.target.value), []);
+
+  const handleChangeJobs = React.useCallback(() => {
```
vilanelle: It's good practice to extract calls to api. We could do so with the use of a custom hook that could be imported from a separate file, to achieve a nice separation of concerns:

```
const useFetchJobs = (url: string) => {
  const [isFetching, setIsFetching] = useState(false);
  const [jobPosts, setJobPosts] = useState<Job[]>([]);

  useEffect(() => {
    if (!url) return;
    const fetchData = async () => {
      setIsFetching(true);
      const response = await fetch(url);
      const data = await response.json();
      setJobPosts(data);
      setIsFetching(false);
    };

    fetchData();
  }, [url]);

  return { isFetching, jobPosts };
};
```

Then we can use the return values from the hook in our component (replacing current line 36 and 39):
```
const { jobPosts, isFetching } = useFetchJobs(url); 
```
If you like the idea, there's more about custom hooks here, including some guidelines on how to handle errors: https://www.smashingmagazine.com/2020/07/custom-react-hook-fetch-cache-data/

vilanelle (changes requested):

src/detail-page/components/TabJobContent/FetchJobContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+            url: job.url,
+            key: job.id,
+          }
+          )),
+        );
+        setFetching(false);
+      }).catch(() => {
+        <Redirect to="/404" />;
+      });
+    console.log(jobPosts);
```
Stajgar: Do we need this `console.log`?

src/detail-page/components/TabJobContent/FetchJobElement.tsx
Stajgar: This import seems redundant (as far as I can see we are not using anything from the `react` library in this component).

joannajerz: Yes, but I need this because i use jsx here ('React' must be in scope when using JSX - i get this message)

vilanelle: Actually, we no longer need to have React in scope :) we have a new TypeScript version which enables us to use JSX without explicitly importing React.

You can read more about it here: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

In short, we have Typescript 4.1+ in the project, meaning we get new JSX transform by default. You can check it in tsconfig file - the "jsx" prop is now "jsx-react" which means we have the new JSX transform:
image

The only thing we should do, is update our linter configuration to handle new JSX transform and then don't have to import React: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#eslint

src/data/ITProfessions.ts
joannajerz: done

src/detail-page/components/JobsTab.tsx
joannajerz: Done

src/detail-page/components/JobsTab.tsx
joannajerz: done

src/detail-page/components/TabJobContent/FetchJobContent.tsx
joannajerz: done

src/detail-page/components/TabJobContent/FetchJobContent.tsx
joannajerz: done

src/detail-page/components/TabJobContent/FetchJobElement.tsx
joannajerz: done

src/detail-page/components/TabJobContent/FetchJobContent.tsx
joannajerz: done

src/detail-page/components/TabJobContent/FetchJobElement.tsx
joannajerz: Thanks;)

src/detail-page/components/TabJobContent/FetchJobElement.tsx
joannajerz: done;)

src/detail-page/components/TabJobContent/FetchJobContent.tsx
joannajerz: done

src/detail-page/components/TabJobContent/FetchJobContent.tsx
joannajerz: I'm not 100% sure it is correct, but I use it to callback function (This hook is used to memoize functions, only for performance). When functions are passed to other components this hooks ,,limits"  us initializing the function every time the parent component is re-rendered.

vilanelle: It is often hard to tell when to use performance optimization. Many experienced programmers recommend, as a rule of thumb, not doing any optimization until the app begins to slow down. 
But there are some cases when we will want to prevent future slowdown before it starts to show. These cases will usually concern preventing unneeded http requests or heavy computation (the latter not very likely in our case). 
In the case of `handleChangeRole/City` we don't have an http request or heavy computation, so I think we don't really need `useCallback`.
This article explains pretty well (in my opinion :) ) when we can try to optimize and when we shouldn't: https://dmitripavlutin.com/dont-overuse-react-usecallback/

src/detail-page/components/TabJobContent/useFetchJobs.ts
```diff
@@ -0,0 +1,33 @@
+import React, { useState, useEffect } from 'react';
+import Job from '../../../data/Jobs';
+
+const useFetchJobs = (url: string) => {
```
vilanelle: It's great that we have extracted the fetch logic to a separate hook! :)

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const URL = `https://cors.bridged.cc/https://jobs.github.com/positions.json?title=${role}&location=${city}`;
+  const { jobPosts, isFetching } = useFetchJobs(URL);
+  const handleChangeRole = React.useCallback(e => setRole(e.target.value), []);
+  const handleChangeCity = React.useCallback(e => setCity(e.target.value), []);
+
+  return (
+    <>
+      <SearchInput type="search" placeholder="Job title" value={role} disableUnderline onChange={handleChangeRole} />
+      <SearchInput placeholder="City" disableUnderline value={city} onChange={handleChangeCity} />
+      <SearchButton color="secondary" variant="contained" onClick={useFetchJobs(URL)}>Search</SearchButton>
```
vilanelle: Unfortunately, we can't call a React hook from here, because we'd be breaking [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level).

We need a bit of mental gymnastics here and think about it the hooks way. Specifically: we can't call a hook directly via onClick. Hooks fire on changes to hooks deps array, in our case this array includes `url` value:

image

If the hook will fire on change to url, all we have to do is update url on click. Instead of hardcoding url as a string, let's set it via `setState` and update state on click:
```
const [url, setUrl] = useState('');
...
<SearchButton color="secondary" variant="contained" onClick={() => setUrl(`https://cors.bridged.cc/https://jobs.github.com/positions.json?title=${role}&location=${city}`)}>Search</SearchButton>

Comment (joannajerz) [reply src/detail-page/components/TabJobContent/JobTab.tsx]: I tried to do it this way, but it doesn't seem to work. Link changes after clicking, but the api stuff is not being uploaded to jobPosts. The link also only changes when I double-click.

Comment (joannajerz) [reply src/detail-page/components/TabJobContent/JobTab.tsx]: done;)

File: src/detail-page/components/TabJobContent/JobTab.tsx [hunk 20]
```diff
@@ -0,0 +1,64 @@
...
+import useFetchJobs from './useFetchJobs';
+import Job from '../../../data/Jobs';
+import hiringPicture from '../../../icons/hiring.svg';
+
+const SearchInput = styled(Input)`
+  width: 250px;
+  height: 35px;
+  border: 1px #CFCFCF solid;
+  margin: ${props => props.theme.spacing(5)}px 0 ${props => props.theme.spacing(1.5)}px ${props => props.theme.spacing(1.5)}px ;
+  padding-left: 0 ${props => props.theme.spacing(1.5)}px
```
klaraborowska: I think it would look nice, if we added some padding on the left to the input fields.

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <SearchButton
+        color="secondary"
+        variant="contained"
+        onClick={() => {
+          setUrl(`https://cors.bridged.cc/https://jobs.github.com/positions.json?title=${role}&location=${city}`);
+        }}
+      >
+        Search
+      </SearchButton>
+      <HiringImage data={hiringPicture} />
```
klaraborowska: I'm not sure, if this was a part of this task, but I think we should not display this image, when we have job list. There is not enough space for both.  What do you think about it?

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const [city, setCity] = React.useState('');
+  const [url, setUrl] = React.useState('');
+  const handleChangeRole = React.useCallback(e => setRole(e.target.value), []);
+  const handleChangeCity = React.useCallback(e => setCity(e.target.value), []);
+  const { jobPosts, isFetching } = useFetchJobs(url);
+
+  return (
+    <>
+      <SearchInput type="search" placeholder="Job title" value={role} disableUnderline onChange={handleChangeRole} />
+      <SearchInput placeholder="City" disableUnderline value={city} onChange={handleChangeCity} />
```
klaraborowska: It would be nice to add a small UX improvement - when you type something in input field and press enter, it could invoke the search acton.

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <SearchButton
+        color="secondary"
+        variant="contained"
+        onClick={() => {
+          setUrl(`https://cors.bridged.cc/https://jobs.github.com/positions.json?title=${role}&location=${city}`);
```
klaraborowska: I think that clicking on search button could clear the  `jobPosts` array, so we see spinner only and than new jobs list, when they are fetched. For now we have both - old list and a spinner. If there was an error, when fetching new data, it could be confusing, cause you would still see the old list.

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -0,0 +1,64 @@
...
+  width: 100px;
+  margin: ${props => props.theme.spacing(1.5)}px;
+  display: block;
+  color: white;
+`;
+
+const HiringImage = styled('object')`
+  width: 250px;
+  float: right;
+  margin: 0 ${props => props.theme.spacing(10)}px ${props => props.theme.spacing(10)}px 0;
```
klaraborowska: Positioning an element using big margin value is not the best solution because there is a big chance that you will have some problems with that in the future, or at least you will have to overwrite it for smaller screens. Here the value is 80px so for example for mobile view it's way to big. I would probably try to use flex-box here and align the element vertically and horizontally.

src/detail-page/components/TabJobContent/JobListItem.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <li key={job.id}>
+      <h4>{job.company} {job.title} </h4>
+      {job.description}
+      <Button variant="contained" href={job.url}>See More</Button>
+    </li>
+  );
+};
+
+export default JobListItem;
```
vilanelle: Let's switch to named exports which give us greater control and clarity over what we import. Also, they will not generate weird linter errors :D 
```
export JobListItem;
...
import { JobListItem } from ...

File: src/detail-page/components/TabJobContent/useFetchJobs.ts [hunk 26]
```diff
@@ -0,0 +1,24 @@
...
+      setIsFetching(false);
+    };
+
+    fetchData();
+  }, [url]);
+
+  return { isFetching, jobPosts };
+};
+
+export default useFetchJobs;
```
vilanelle: Let's use named exports.

vilanelle (approved):

src/detail-page/components/TabJobContent/JobTab.tsx
joannajerz: done

src/detail-page/components/TabJobContent/JobTab.tsx
joannajerz: done

src/detail-page/components/TabJobContent/JobTab.tsx
joannajerz: done

src/detail-page/components/TabJobContent/JobTab.tsx
joannajerz: done

src/detail-page/components/TabJobContent/JobListItem.tsx
joannajerz: done

src/detail-page/components/TabJobContent/useFetchJobs.ts
joannajerz: done

src/layout/theme.ts
```diff
@@ -53,10 +53,14 @@ export const theme = createMuiTheme({
     primary: {
       main: '#725ec0',
     },
+    secondary: {
```
vilanelle, line 63: 👍 for managing app colors through our theme! 👌 One small change I would do is add the new color as `primary.dark` rather than `secondary.main`. 
Primary and secondary colors in a palette are usually contrast colors, you can check it out in this example: https://material-ui.com/customization/palette/

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -0,0 +1,88 @@
...
+`;
+
+const SearchButton = styled(Button)`
+  width: 100px;
+  margin: ${props => props.theme.spacing(1.5)}px;
+  display: block;
+  color: white;
+`;
+
+const HiringImage = styled('object')`
```
vilanelle: I wonder if we need to use `object` tag here. After all, all we need is to display a static svg. Maybe `img` would be a better choice here?

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const [city, setCity] = React.useState('');
+  const [url, setUrl] = React.useState('');
+  const handleChangeRole = React.useCallback(e => setRole(e.target.value), []);
+  const handleChangeCity = React.useCallback(e => setCity(e.target.value), []);
+  /* eslint-disable prefer-const */
+  let { jobPosts, isFetching } = useFetchJobs(url);
+
+  const handleSearch = () => {
+    jobPosts = [];
+    setUrl(`https://cors.bridged.cc/https://jobs.github.com/positions.json?title=${role}&location=${city}`);
```
vilanelle: I checked github jobs api and it looks like it doesn't accept `title` param. I guess we should replace it with `description` to improve our search results :) We could also reflect it in placeholder (use `Description` instead of `Job title`). In the future, if we decide to use multiple apis we can change it.

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -0,0 +1,88 @@
...
+  padding:0;
+`;
+const JobTab: React.FC = () => {
+  const [role, setRole] = React.useState('');
+  const [city, setCity] = React.useState('');
+  const [url, setUrl] = React.useState('');
+  const handleChangeRole = React.useCallback(e => setRole(e.target.value), []);
+  const handleChangeCity = React.useCallback(e => setCity(e.target.value), []);
+  /* eslint-disable prefer-const */
+  let { jobPosts, isFetching } = useFetchJobs(url);
```
vilanelle: Linter complaints often signal that something's not quite ideal in our code :) In this case, it's probably not the best idea to overwrite the value returned from the hook by setting `jobPosts` to `[]` - it creates two sources of truth for the data and that's never good. But we can easily avoid it by resetting the value directly in the hook, the place responsible for managing this data:
```
  useEffect(() => {
    if (!url) {
      return;
    }
    setJobPosts([]);
    const fetchData = async () => {
...
```

src/detail-page/components/TabJobContent/useFetchJobs.ts
```diff
@@ -0,0 +1,22 @@
+import React, { useState, useEffect } from 'react';
```
vilanelle, line 1: It's great that we have extracted the job fetching logic to a separate part of code! 💪

src/detail-page/components/TabJobContent/JobListItem.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export const JobListItem: React.FC<Props> = ({ job }) => {
+  return (
+
+    <JobListElement key={job.id}>
+      <JobListColorElement> </JobListColorElement>
+      <JobListContent>
+        <h4>{job.company} {job.title} </h4>
+        <Typography variant="body2" color="textSecondary">{job.description.replace(/<\/?[^>]+(>|$)/g, '').substring(0, 200)}...</Typography>
```
vilanelle: What do you think about extracting the formatting logic to a named function, for example `formatResponse`? I'd even go a step further and use a named variable for the regex, sth like:

```
const matchHtmlTag = /<\/?[^>]+(>|$)/g;
const formatDescription = (description: string) => description.replace(matchHtmlTag, '').substring(0, 200);
```

Regex expressions are notorious for being difficult to understand, so the above should be helpful to any future developers (or ourselves in a few months' time ;) ).

vilanelle (approved):

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -0,0 +1,87 @@
...
+const JobList = styled('ul')`
+  padding:0;
+`;
+const JobTab: React.FC = () => {
+  const [role, setRole] = React.useState('');
+  const [city, setCity] = React.useState('');
+  const [url, setUrl] = React.useState('');
+  const handleChangeRole = React.useCallback(e => setRole(e.target.value), []);
+  const handleChangeCity = React.useCallback(e => setCity(e.target.value), []);
+  /* eslint-disable prefer-const */
```
vilanelle: We are no longer changing the value of `jobPosts` so we can replace `let with `const` and remove eslint-disable 👌

vilanelle (approved):

src/detail-page/components/TabJobContent/JobTab.tsx
joannajerz: done;)

OlgaMilczek (approved):

PR #18, opened 2021-03-28
Drm 62 - add new date structure to project.

OlgaMilczek:
https://dareit.atlassian.net/browse/DRM-61

- Add new date structure to project. 
- Add interfaces for datas and empty files for data. 
- Add headlining long descriptions in DetailPage Profession Card. 

Changes not related to the tag itself:
- Change place of back button according to mockups. 
-


src/data/detail-page/Descriptions/devopsDescription.ts
```diff
@@ -0,0 +1,4 @@
+export const description = 'A DevOps engineers combine knowledge about application coding, '
```
vilanelle: To avoid the + before new lines we can use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here. They allow for multiline strings.

src/data/detail-page/Resources/ResourcesInterfaces.ts
```diff
@@ -0,0 +1,39 @@
+export interface Book {
```
vilanelle: I'm wondering if we are not complicating things too much - do we need our resource types to be different? For example, I'm not sure if we need to define blog | portal | podcast type, because it won't really matter for displaying the resource.

At first glance, for displaying the resource on the list we should be fine with title, description and url:
```
interface Resource {
   title: string;
   description: string;
   url: string;
}
```

Things like language or author could be easily included in description when needed. In the future, if we needed to add a prop specific to only one type of resource, [we could extend our base interface ](https://www.typescriptlang.org/docs/handbook/2/objects.html#extending-types)like so:
```
interface PaidResource extends Resource {
   price: string;
}
```

src/data/detail-page/Resources/ResourcesInterfaces.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface PortalOrPodcast {
+    type: 'portal' | 'podcast' | 'blog';
+    description?: string;
+    language?: string;
+    link: string;
+}
+
+export interface Resources {
+    books?: Book[];
```
vilanelle: Optional properties are a so-called [code smell ](https://en.wikipedia.org/wiki/Code_smell) and should be avoided if possible. Here, there's no need to mark them as optional because they are all arrays and arrays can be empty 🙃 So let's remove the optional markers (`?`) for all props in the Resources interface.

src/detail-page/DetailPage.tsx
```diff
@@ -20,30 +18,38 @@ export const TabContainer = styled('div')`
...
+    if (professionId) {
+      import(`../data/detail-page/Descriptions/${professionId}Description.ts`)
+        .then(response => setDetailDescription(response.description));
+    }
+  }, [professionId]);
+
   return (
     <>
-      {profession && (
+      {(profession && detailDescription) ? (
```
vilanelle, line 51: 👍

src/data/detail-page/Graphs/GraphInterfaces.ts
```diff
@@ -0,0 +1,5 @@
+export interface Graph {
```
vilanelle: I think it's a `Chart` not `Graph` 🤔I could be wrong, but maybe we should check it.

src/data/detail-page/Descriptions/backendDescription.ts
```diff
@@ -0,0 +1,2 @@
+// TODO: Add longer description
```
vilanelle: Let's ensure we have uniform file and directory names across the project. 
Directory names should follow kebab-case (as in the rest of the project). Files like GraphInterfaces or LearnInterfaces are not jsx components, so they should be camelCase. Project manager is two words and the file name should reflect it: `projectManagerGraph.ts`.
Also, if we our directory name is e.g. descriptions, mayeb we don't need to use the `description` suffix for all the files inside this directory (that is instead of `backendDescription` have simply `backend)`? This would make the directory content far more readable! Finally, it's good practice to ensure the exported const name matches the file name. So in a file `backend.ts` we'd have `export const backend = ...`.
image

vilanelle (changes requested):

src/detail-page/DetailPage.tsx
```diff
@@ -20,30 +18,38 @@ export const TabContainer = styled('div')`
...
         <MainContainer>
+          <BackButton />
```
vilanelle, line 53: Good job with taking the time to adjust our views to the mockups! 👏

src/detail-page/DetailPage.tsx
```diff
@@ -20,30 +18,38 @@ export const TabContainer = styled('div')`
 `;
 
 export const DetailPage: React.FC<{}> = () => {
+  const [detailDescription, setDetailDescription] = useState<string>('');
+
   const { professionId } = useParams<{ professionId: string }>();
   const profession = ITProfessions.find(p => p.id === professionId);
+
+  useEffect(() => {
+    if (professionId) {
```
vilanelle: What will happen if the module can't be found, e.g. if we make a typo in the import path? It may be a good idea to add error handling:
```
 import(`../data/detail-page/Descriptions/${professionId}Description.ts`)
        .then(response => setDetailDescription(response.description))
        .catch(err => handle error... )

File: src/detail-page/components/tabs/Tabs.tsx [hunk 9]
```diff
@@ -1,15 +1,16 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect, Dispatch, SetStateAction } from 'react';
```
vilanelle: I think we have two unused imports here?

src/detail-page/components/tabs/Tabs.tsx
```diff
@@ -50,10 +50,20 @@ const getTabContent = (categoryName: String) => {
...
+export const Tabs: React.FC<TabProps> = ({ professionId }) => {
   const [activeTab, setActiveTab] = useState(0);
+  const [categories, setCategories] = useState<Category[]>([]);
+
   const { path, url } = useRouteMatch();
   const history = useHistory();
+
+  useEffect(() => {
+    if (categories.length === 0) {
+        import(`../../../data/detail-page/Tabs/${professionId}Tabs.ts`)
```
vilanelle: These dynamic imports work beautifully :) however, I dug a bit into documentation and I'm wondering if we should stick with them. Official docs list a number of reasons for using dynamic imports: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports

But our case doesn't really fall into any of these categories. So I'm thinking that it might better to use a static import because it looks like it's best practice for this scenario. What do you think?

If you agree, I propose the following solution:

- rename `categories` consts in individual data files to make the names unique (instead of `catgories` make it `frontend`, `backend`, `uxui` etc.)
- [add barrel exports](https://basarat.gitbook.io/typescript/main-1/barrel): in Tabs directory add a file `index.ts` and export the content of the directory files like so:
```
export * from './devopsTabs';
export * from './frontendTabs';
...
```
- import categories heere using named imports:
```
import { frontend, devops, backend ... } from 'data/detail-page';
```
- set category in useEffect

vilanelle (changes requested):

src/data/detail-page/Descriptions/devopsDescription.ts
OlgaMilczek: Done

src/data/detail-page/Resources/ResourcesInterfaces.ts
OlgaMilczek: Done

src/data/detail-page/Resources/ResourcesInterfaces.ts
OlgaMilczek: Done

src/data/detail-page/Graphs/GraphInterfaces.ts
OlgaMilczek: Yes, they are Charts :) I change that :)

src/data/detail-page/Descriptions/backendDescription.ts
OlgaMilczek: Problem is that I import file by by for component by it route Id. Route id for project menager, devOps are lover case, so they would not import in components if they name will be camel case.

src/detail-page/DetailPage.tsx
OlgaMilczek: I can redirect to our 404 page. :)

src/detail-page/components/tabs/Tabs.tsx
OlgaMilczek: Ok. Delete. :)

src/detail-page/components/tabs/Tabs.tsx
OlgaMilczek: If we use static import then component will have data that they don't need, that's why I suggest to use dynamic import. But this files are not so big, and it doesn't affects performance so we can change it to static import.

src/detail-page/DetailPage.tsx
vilanelle: It's no longer applicable, because we rely on normal imports now :) so we don't have to worry about it :)

src/data/detail-page/charts/backend.ts
```diff
@@ -0,0 +1,5 @@
+import { Chart } from './chartInterfaces';
+
+export const backend: Chart = {
+  hasChart: false,
```
klaraborowska: I'm not sure, if we need the property `hasChart`. Category, for which we don't want to display chart, could have type set to null and data could be an empty array and it would be consistant to other data structures in our app, like for example resources.

src/data/detail-page/tabs/backend.ts
```diff
@@ -0,0 +1,19 @@
...
+  {
+    id: 'jobs',
+    name: 'jobs',
+    text: 'Jobs',
+  },
+  {
+    id: 'resources',
+    name: 'resources',
+    text: 'Useful Resources',
+  },
```
klaraborowska: I have one concern here - we probably want to have the same display name, text etc. for each tab, doesn't matter in which category we are. So I would rather define in here, which tabs we need for a category, in an array with strings, for example:
`const backend = ['learn', 'jobs', 'resources']` 
And we could have one big array with objects for all of the tabs in one place.
 The problem that I see here - imagine that we decide to change the name of resources from `'Useful Resources'` to` 'Best Resources'`. In this solution we need to change it in 6 places and we would probably prefer to change it in one place only :)

vilanelle (approved):

src/data/detail-page/charts/backend.ts
OlgaMilczek: Ok :) Deleted.

src/data/detail-page/tabs/backend.ts
OlgaMilczek: Done :)

klaraborowska (approved):

PR #19, opened 2021-03-29
DRM-67 Added frontend chart to detail page

Stajgar:
https://dareit.atlassian.net/browse/DRM-67


src/detail-page/components/FrontendLearnChart.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import { Bar } from 'react-chartjs-2';
+
+const FrontendLearnChartData = {
+  labels: [
+    'React',
```
vilanelle: I like it that we have separated our data from our chart! 👏 How about we take a step further and make sure our labels always match the figures? I'm thinking specifically of putting labels and their respective percentages into a separate array:
```
const data: ChartDataItem[] = [
   {
      label: 'React',
      value: 75,
   },
....
];
```
Then, in `FrontendLearnChartData`, we would generate `labels`: `data.map(item => item.label);` and similarly the `data` prop in `dataset`. The benefit is that we now have a guarantee that the label will match its value 👌

src/detail-page/components/FrontendLearnChart.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import { Bar } from 'react-chartjs-2';
+
+const FrontendLearnChartData = {
```
vilanelle: This is not a jsx components so we should probably use camelCase for its name :)

src/detail-page/components/FrontendLearnChart.tsx
```diff
@@ -0,0 +1,62 @@
...
+      },
+    ],
+  },
+  title: {
+    display: true,
+    text: 'Most popular JS frameworks in 2020',
+  },
+  legend: {
+    display: false,
+    fontFamily: 'Segoe UI',
```
vilanelle: If we have `display: false` we probably don't need to specify font family :)

src/detail-page/DetailPage.tsx
```diff
@@ -19,6 +20,15 @@ export const TabContainer = styled('div')`
   }
 `;
 
+const PROFESSION_CHARTS = {
+  frontend: <FrontendLearnChart />,
+  backend: null,
+  uxui: null,
+  qa: null,
+  devops: null,
+  projectmanager: null,
```
vilanelle: "Project manager" is two words - could we reflect it and use camelCase? I mean `projectManager`. Not sure about `uxUi`, what do you think? 🤔

src/detail-page/components/FrontendLearnChart.tsx
```diff
@@ -0,0 +1,62 @@
...
+      {
+        gridLines: {
+          display: false,
+        },
+      },
+    ],
+  },
+  title: {
+    display: true,
+    text: 'Most popular JS frameworks in 2020',
```
vilanelle, line 79: We have a nice descriptive title for our chart 💪 Let's also add data source (State of JS, if I'm not mistaken? :) ) and information that our values are percentages to make the chart even more user-friendly!

src/detail-page/DetailPage.tsx
```diff
@@ -37,10 +47,8 @@ export const DetailPage: React.FC<{}> = () => {
           />
           <TabContainer>
             <BackButton />
-            <Tabs
-              categories={profession.categories}
-              profession={profession}
-            />
+            <Tabs categories={profession.categories} profession={profession} />
+            {PROFESSION_CHARTS[profession.id]}
```
vilanelle: Hm, it looks like the chart is outside our tabs component:

image

Let's move it (with the corresponding logic) to LearnTab!

vilanelle (changes requested):

src/detail-page/components/FrontendLearnChart.tsx
```diff
@@ -0,0 +1,62 @@
...
+  },
+  legend: {
+    display: false,
+    fontFamily: 'Segoe UI',
+  },
+};
+export const FrontendLearnChart: React.FC<{}> = () => (
+  <div style={{ height: '200px', width: '400px' }}>
```
klaraborowska: You could go here for maxWidth - then it would scale nicely on smaller screen. But it's just a minor change :)

src/detail-page/DetailPage.tsx
Stajgar: Moved 👍

src/detail-page/DetailPage.tsx
Stajgar: I would gladly do so, but I think those names need to math those from `../../data/home-page/ITProfessions` ?

src/detail-page/components/FrontendLearnChart.tsx
Stajgar: Ok, done

src/detail-page/components/FrontendLearnChart.tsx
Stajgar: Oh, of course. Done

src/detail-page/components/FrontendLearnChart.tsx
Stajgar: Done

src/detail-page/components/FrontendLearnChart.tsx
Stajgar: Good idea. It's done :)

src/detail-page/components/FrontendLearnChart.tsx
Stajgar: Good catch. I've removed it

src/detail-page/components/LearnTab.tsx
```diff
@@ -20,6 +30,7 @@ export const LearnTab: React.FC<LearnProps> = ({ professionId }) => (
     <LearnList professionId={professionId} />
     <div>
       {/* Should be replace with actual content */}
```
vilanelle, line 32: I don't think we need this comment anymore :)

PR #20, opened 2021-04-12
Drm 44 - What to learn list

OlgaMilczek:
[JIRA TASK](https://dareit.atlassian.net/browse/DRM-43)

Create what to learn list and add roadmap buttons.


src/detail-page/components/LearnList.tsx
```diff
@@ -0,0 +1,73 @@
...
+    padding-left: 1.5rem;
+`;
+
+const StyledLi = styled('li')`
+    margin: .8rem 0;
+    font-size: .95rem;
+    position: relative;
+
+    &::first-letter {
+        text-transform: uppercase;
```
vilanelle: It's not working on firefox :/

src/detail-page/components/LearnList.tsx
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    &::before {
```
vilanelle: [Optional] We could use a Material icon instead of custom css.

src/detail-page/components/LearnTab.tsx
```diff
@@ -1,12 +1,25 @@
...
+  display: grid;
+  grid-template-columns: repeat(auto-fit, minmax(min(400px, 100%), 1fr));
+  grid-gap: 2em;
+  padding: 1.5em;
+  min-height: 50vh;
+`;
 
 interface LearnProps {
-  content: string;
+  professionId: ProfessionKey;
```
vilanelle: [optional] Maybe we could unify names and use only one of the two (id vs key)?

src/detail-page/hooks/useLearnList.tsx
```diff
@@ -0,0 +1,22 @@
+import { backend, devOps, frontend, projectManager, qa, uxUi } from '../../data/detail-page/learn/index';
+import { Learn } from '../../data/detail-page/learn/LearnInterfaces';
+import { ProfessionKey } from '../../data/home-page/ITProfessions';
+
+export const getData = (professionId: ProfessionKey): Learn => {
```
vilanelle: Let's unify the file name and the exported function name. Currently, one is named `useLearnList` and the other `getData` which may be misleading. Also, just in case: if we decide to rename the function to `useLearnList`, then it should be a hook and it should follow the [rules of hooks](https://reactjs.org/docs/hooks-rules.html#gatsby-focus-wrapper).

src/data/detail-page/tabs/backend.ts
```diff
@@ -0,0 +1,3 @@
+import { categoryId } from './tabsInterfaces';
+
+export const backend: categoryId[] = ['learn', 'jobs', 'resources'];
```
vilanelle: Types and interfaces names should be capitalized by convention: https://google.github.io/styleguide/tsguide.html#identifiers 

Let's rename our type to `CategoryId` :)

src/data/detail-page/tabs/tabsInterfaces.ts
```diff
@@ -0,0 +1,11 @@
+export interface Category {
+    id: categoryId;
+    name: string;
+    text: string;
+}
+
+export type categoryId = 'learn' | 'jobs' | 'resources'
+
+export type categoriesModel = {
```
vilanelle: Also here, let's rename the type to `CategoriesModel`

src/data/detail-page/tabs/tabsInterfaces.ts
```diff
@@ -0,0 +1,11 @@
...
+    [key in categoryId]: Category
```
vilanelle: 👌

src/detail-page/DetailPage.tsx
```diff
@@ -20,27 +19,48 @@ export const TabContainer = styled('div')`
   }
 `;
 
+const getData = (professionId: ProfessionKey) => {
```
vilanelle: Could we add return type for this function?

src/detail-page/components/LearnList.tsx
```diff
@@ -0,0 +1,73 @@
...
+        transform: translateY(6px)
+    }
+`;
+
+export const LearnList: React.FC<LearnListProps> = ({ professionId }) => {
+  const data = getData(professionId);
+  return (
+    <StyledDiv>
+      <StyledList>
+        {data.list.map((item, index) => (
```
vilanelle, line 61: We are not using index, so we don't have to declare it as param :)

src/detail-page/hooks/useTabData.tsx
```diff
@@ -0,0 +1,60 @@
...
+    case 'qa':
+      return qa;
+    case 'uxui':
+      return uxUi;
+    default:
+      return [];
+  }
+};
+
+export const useTabData = (professionId: ProfessionKey) => {
```
vilanelle: What a great custom hook! 🚀We now have our tab switching logic nicely encapsulated in one place! Excellent :)

src/detail-page/hooks/useTabData.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const [activeTab, setActiveTab] = useState(0);
+  const categories = getData(professionId);
+
+  const { path, url } = useRouteMatch();
```
vilanelle: We're not using path or url anywhere inside the hook. Maybe we can simply remove it from the hook and use `useRouteMatch` directly in the components?

vilanelle (changes requested):

src/detail-page/components/LearnList.tsx
OlgaMilczek: Oh... that's weird.
https://caniuse.com/?search=%3A%3Afirst-letter - but actually, They write that it can be something wrong on Firefox. (I didn't check this tab)
So I can change it to JS solution, to be sure that will work for all browsers. :)

src/detail-page/hooks/useTabData.tsx
OlgaMilczek: We use url in `handleTabClicked` function, but we can also pass it as an argument ;)

src/detail-page/components/LearnList.tsx
OlgaMilczek: Yes, I want to use it as a key, but our linter doesn't want to cooperate so i use item instead, but maybe I will fight with it, and change key for index, I think it will be a little bit better in this case.

src/detail-page/DetailPage.tsx
OlgaMilczek: Done

src/data/detail-page/tabs/backend.ts
OlgaMilczek: Done

src/data/detail-page/tabs/tabsInterfaces.ts
OlgaMilczek: Done :)

src/detail-page/components/LearnTab.tsx
OlgaMilczek: We can change it for Id.

src/detail-page/hooks/useLearnList.tsx
OlgaMilczek: I was thinking maybe I will need some hook here, but in the end i need only this one function. But you are right it is not a hook, so it should not be here :D So I will move it  to components helpers.

src/detail-page/components/LearnList.tsx
OlgaMilczek: Done :)

src/detail-page/components/tabs/TabPanel.tsx
```diff
@@ -1,19 +1,25 @@
...
   getTabContent: (categoryId: string) => ReactElement;
   syncTabWithPath: (categoryId: string) => void;
 }
 
+const StyledDiv = styled('div')`
+    min-height: 62vh;
+    display: flex;
+`;
+
 export const TabPanel: React.FC<Props> = ({ getTabContent, syncTabWithPath }) => {
```
ireshka: Other components use convention `ComponentNameProps` eg. `TabProps`, `LearnListProps`. Maybe we can change here name for `TabPanelProps`?

src/detail-page/hooks/useTabData.tsx
```diff
@@ -0,0 +1,59 @@
...
+      case 'jobs':
+        return <JobsTab professionId={professionId} />;
+      case 'learn':
+        return <LearnTab professionId={professionId} />;
+      case 'resources':
+        return <ResourcesTab professionId={professionId} />;
+      default:
+        return <Redirect to="/404" />;
+    }
+  };
```
ireshka: In `data/detail-page/tabs/tabsInterfaces` you created `CategoryId` type. Maybe we could use it here instead of `String` (`categoryName: CategoryId`). If we decide on this we should also change type in `TabPanel` component props (`getTabContent`)

src/data/detail-page/tabs/backend.ts
```diff
@@ -1,3 +1,3 @@
-import { categoryId } from './tabsInterfaces';
+import { CategoryId } from './tabsInterfaces';
 
-export const backend: categoryId[] = ['learn', 'jobs', 'resources'];
+export const backend: CategoryId[] = ['learn', 'jobs', 'resources'];
```
ireshka, line 3: For futher discussion: I see that all category files from tabs directory have identical content. Do we need seperated files with categories for future improvements?

ireshka (approved):

src/detail-page/hooks/useTabData.tsx
vilanelle: Oh, my bad, I missed it :) we can leave it here, then.

.eslintrc.json
```diff
@@ -55,24 +71,33 @@
...
+      {
+        "SwitchCase": 1
+      }
     ],
     "react/jsx-props-no-spreading": "off",
     "react/jsx-one-expression-per-line": "off",
     "react/prop-types": "off",
     "react-hooks/rules-of-hooks": "error",
     "react-hooks/exhaustive-deps": "warn",
+    "react/no-array-index-key": "off",
```
vilanelle: Danger alert! ;) We should not disable this rule. React expressly discourages the use of indexes as key and for good reasons: to name just one, it may lead to unpredictable behavior, e.g. components failing to re-render. There may be some exceptions to this rule, meaning that we may sometimes have an index as a key, but we should avoid it (and keep the lint rule to remind us about it).
You can read more about the importance of keys and why we shouldn't use indexes for keys [in the official docs](https://reactjs.org/docs/lists-and-keys.html#keys) and [a recommended article](https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318).

src/detail-page/components/LearnList.tsx
vilanelle: As mentioned in my other comment, we should avoid using indexes as keys as a rule of thumb. However, in this case it seems safe to do: we have a static list whose order / length will never change. Also, we don't have any item ids that we could use instead of indexes and it seems a bit of an overkill to generate ids for list items. 
However, we should definitely not disable the respective eslint rule globally, because we want to be warned in future cases. Instead let's disable it locally:
```
<StyledList>
        {/* eslint-disable react/no-array-index-key */}
        {data.list.map((item, index) => (
          <StyledLi key={index}>
            <Bullet color="primary" />
           ...
</StyledList>

File: src/detail-page/components/tabs/TabPanel.tsx (hunk 12 above)
Comment (vilanelle) [reply src/detail-page/components/tabs/TabPanel.tsx]: It's a very interesting question! :) Actually, naming of variables is one of the most difficult things in programming and we often have a dilemma whether to choose a short and easy to read variable name or rather a long and less readable, but more descriptive name.
If in doubt, it is helpful to think about the variable scope and context. For a small scope (e.g. a local function) we can use a shorter name. If it's broader, global, the name will probably have to be longer and more descriptive.
So in this case, we are dealing with a relatively small scope of a single component (the variable is not exported). Also, the context is quite clear - component props are a very common part of React code. So my subjective opinion is that we are quite fine with `Props` here.

File: src/detail-page/components/LearnList.tsx [hunk 16]
```diff
@@ -0,0 +1,79 @@
...
+    margin-top: 0;
+    margin-bottom: 1rem;
+    padding-left: 0;
+`;
+
+const StyledLi = styled('li')`
+    display: flex;
+    line-height: 1.2;
+
+    &:not(:first-child) {
```
vilanelle: We can achieve the same effect with this css:
```
margin: 0 0 .8rem;
```
It reduces complexity a bit, by not using `first:child`. Also we no longer have to rely on [margin collapsing](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing)  :)

src/detail-page/components/LearnList.tsx
```diff
@@ -0,0 +1,79 @@
...
+      margin: .8rem 0;
+    }
+
+    p {
```
vilanelle: We can extract the `p` as a separate styled component in order to reduce css nesting (and the related complexity) as much as possible! :muscle:

vilanelle (approved):

src/detail-page/components/LearnList.tsx
OlgaMilczek: Done :)

src/detail-page/hooks/useTabData.tsx
OlgaMilczek: Done :)

src/detail-page/hooks/useTabData.tsx
OlgaMilczek: Done :)

.eslintrc.json
OlgaMilczek: Done :)

src/detail-page/components/LearnList.tsx
OlgaMilczek: Ok. Done :)

src/detail-page/components/LearnList.tsx
OlgaMilczek: Done :)

src/data/detail-page/tabs/backend.ts
OlgaMilczek: Yes, they are the same, but maybe in the future they we would like to make the different or do note have all of them in some specialization. That's why I make it separate for each one.

PR #21, opened 2021-04-19
DRM-66 -  useful resources tab content

ireshka:
Task: [DRM-66](https://dareit.atlassian.net/browse/DRM-66)


src/index.tsx
```diff
@@ -4,12 +4,7 @@ import './index.css';
 import { App } from './App';
 import reportWebVitals from './reportWebVitals';
 
-ReactDOM.render(
-  <React.StrictMode>
-    <App />
-  </React.StrictMode>,
-  document.getElementById('root'),
-);
+ReactDOM.render(<App />, document.getElementById('root'));
```
vilanelle, line 7: Is there any specific reason why we removed `StrictMode`? It could help us spot potential problems in the app.

src/detail-page/components/accordion/AccordionListItem.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const StyledListElementDescription = styled(Typography)`
+  display: block;
+`;
+
+export const AccordionListItem: React.FC<AccordionListItemProps> = ({
+  listElement,
+}) => {
+  const { url, title, description } = listElement;
+  const hasDescription = url !== null;
```
vilanelle: I like how we name our condition to make the template more readable 👏

src/data/detail-page/resources/ResourcesTypeNames.ts
```diff
@@ -0,0 +1,12 @@
+import { ResourcesType } from './ResourcesInterfaces';
+
+export const ResourcesTypeNames: Record<ResourcesType, string> = {
```
vilanelle: (optional) This looks like a good case for using an [enum](https://blog.logrocket.com/writing-readable-code-with-typescript-enums-a84864f340e9/) :)
```
export enum ResourcesTypeNames {
  books = 'Books',
  courses = 'Online courses',
  platforms = 'Platforms',
  tools = 'Useful tools',
  communities = 'Communities',
  portalsAndPodcasts = 'Portals and Podcasts',
  games = 'Games & Coding Challenges',
  varia = 'Varia',
}

File: src/data/detail-page/resources/backend.ts [hunk 4]
```diff
@@ -1,12 +1,51 @@
...
+import { v4 as uuid } from 'uuid';
+import { Resources } from './ResourcesInterfaces';
 
 export const backend: Resources = {
   books: [],
   courses: [],
   platforms: [],
   tools: [],
-  communities: [],
+  communities: [
```
vilanelle, line 9: Thanks for doing all the work needed to add our data! 👏

src/data/detail-page/resources/ResourcesInterfaces.ts
```diff
@@ -0,0 +1,10 @@
+export type ResourcesType = 'books' | 'courses' | 'platforms' | 'tools' | 'communities' | 'portalsAndPodcasts' | 'games' | 'varia';
+
+export interface Resource {
+  title: string;
+  description: string | null;
+  url: string;
+  id: string;
+}
+
+export type Resources = Record<ResourcesType, Resource[]>;
```
vilanelle: :+1:

src/data/detail-page/resources/uxUi.ts
```diff
@@ -1,12 +1,64 @@
-import { Resources } from './resourcesInterfaces';
+import { v4 as uuid } from 'uuid';
+import { Resources } from './ResourcesInterfaces';
 
 export const uxUi: Resources = {
-  books: [],
+  books: [
+    {
+      title: 'Don\'t Make Me Think',
```
vilanelle, line 7: I have to read this one :D

src/detail-page/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { Resource } from '../../../data/detail-page/resources/ResourcesInterfaces';
+
+interface AccordionProps {
+  header: string;
+  list: Resource[];
+  panelNumber: number;
+}
+
+const StyledAccordion = styled(props => <MuiAccordion {...props} />)`
+  border: 1px solid ${({ theme }) => theme.palette.text.disabled};
```
vilanelle: Maybe instead of using a `text` color for the border, we can add `grey` to our theme? We can use the same color.

```  palette: {
    primary: {
      main: '#725ec0',
    },
    secondary: {
      main: '#d99cff',
      contrastText: '#ffffff',
    },
    grey: {
      300: '#22222233',
    },
    text: {
...

File: src/detail-page/components/accordion/Accordion.tsx [hunk 8]
```diff
@@ -0,0 +1,79 @@
...
+import { AccordionListItem } from './AccordionListItem';
+import { Resource } from '../../../data/detail-page/resources/ResourcesInterfaces';
+
+interface AccordionProps {
+  header: string;
+  list: Resource[];
+  panelNumber: number;
+}
+
+const StyledAccordion = styled(props => <MuiAccordion {...props} />)`
```
vilanelle: [curious] Do we need to pass props through this function? I thought styled components preserved props by default?

src/detail-page/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,79 @@
...
+  List,
+  Typography,
+} from '@material-ui/core';
+import { ExpandMore } from '@material-ui/icons';
+import { AccordionListItem } from './AccordionListItem';
+import { Resource } from '../../../data/detail-page/resources/ResourcesInterfaces';
+
+interface AccordionProps {
+  header: string;
+  list: Resource[];
```
vilanelle: Maybe  `resources` instead of `list`? Renaming things in code can become misleading 🤷

src/detail-page/components/ResourcesTab.tsx
```diff
@@ -1,13 +1,42 @@
...
+interface ResourcesTabProps {
   professionId: ProfessionId;
 }
 
-export const ResourcesTab: React.FC<ResourcesProps> = ({ professionId }) => (
-  <div>
-    {/* Should be replace with actual content */}
-    {professionId}
-  </div>
+const TabContainer = styled('div')`
```
vilanelle, line 11: This is definitely not part of your PR, but I'm wondering if we should have a separate container for each tab (Resrouces, Learn, etc.). We'd like them to be nicely consistent, no? Maybe we can discuss it at the next meeting.

src/detail-page/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
```
vilanelle, line 1: It looks great 😍 
image

src/detail-page/components/ResourcesList.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const ResourcesList: React.FC<ResourcesListProps> = ({
+  professionId,
+}) => {
+  const resourcesData = getData(professionId);
+  return (
+    <div>
+      {Object.keys(resourcesData).map((resourceType, index) => {
+        const resourceTitle = ResourcesTypeNames[resourceType as ResourcesType];
+        const areResources = resourcesData[resourceType as ResourcesType].length !== 0;
```
vilanelle: Hm, maybe `areResourcesAvailable`..? `areResources` sounds a bit mysterious :)

vilanelle (changes requested):

src/index.tsx
ireshka: With `strict mode` an app works but there is an error in console according to `findDOMNode is depracted in Strict Mode.` - it's related to `Accordion` component. More info you can find on these issues: https://github.com/mui-org/material-ui/issues/20561 & https://github.com/mui-org/material-ui/issues/13394 . I didn't find another solution for this error except disabling `Strict mode`.
strict-mode

src/data/detail-page/resources/ResourcesTypeNames.ts
ireshka: I decided to use `Record` instead of `enum` because I also use `ResourceType` in `Resources` interface and I wanted to keep consistency between them. Is `enum` better here or it is a matter of choice?

src/detail-page/components/accordion/Accordion.tsx
ireshka: Done 👍🏻

src/detail-page/components/accordion/Accordion.tsx
ireshka: I admit that it is a kind of magic for me a little. :D
In `Accordion` & `AccordionSummary` I added passing props through function when I experimented with other ways of styling (passing `expanded` props instead of classes). Now with class styling I removed it and it looks all right.
But in `StyledListItem` component (which is styled `ListItem` component) when I remove function I get typescript error.
list-item

src/detail-page/components/accordion/Accordion.tsx
ireshka: Done 👍🏻

src/detail-page/components/ResourcesList.tsx
ireshka: Done 👍🏻

src/data/detail-page/resources/ResourcesTypeNames.ts
vilanelle: enums are the go-to solution when we want a predefined list of values, however, you're right, that in this case there is also the issue of keeping the types consistent, so Record is also a good option :)
Just a fun fact for future reference: we could create a type using enum keys (but I'm not sure if this would be more readable than the ResourcesType type that we have now):
```
export enum ResourcesTypeNames {
  books = 'Books',
  courses = 'Online courses',
  platforms = 'Platforms',
  tools = 'Useful tools',
  communities = 'Communities',
  portalsAndPodcasts = 'Portals and Podcasts',
  games = 'Games & Coding Challenges',
  varia = 'Varia'
}

export type ResourcesType = keyof typeof ResourcesTypeNames; // this could replace our ResourcesType type

Lastly, maybe we could rename `ResourcesType` and make it singular (that is `ResourceType`)? Plurals in the names describing singular entities tend to be misleading.

File: src/detail-page/components/LearnTab.tsx [hunk 13]
```diff
@@ -7,7 +7,7 @@ const TabContainer = styled('div')`
   display: grid;
   grid-template-columns: repeat(auto-fit, minmax(min(400px, 100%), 1fr));
   grid-gap: 2em;
-  padding: 1.5em;
+  /* padding: 1.5em; */
```
vilanelle: I guess we can remove this?

src/detail-page/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,80 @@
...
+  &.Mui-expanded + & {
+    border-top: 1px solid ${({ theme }) => theme.palette.grey[300]};
+  }
+
+  &.Mui-expanded {
+    border-top: 1px solid ${({ theme }) => theme.palette.grey[300]};
+  }
+`;
+
+// const StyledAccordionSummary = styled(props => <MuiAccordionSummary {...props} />)`
```
vilanelle: We can remove it now? :)

src/index.tsx
vilanelle: I see, that's not ideal :/ I guess there's little we can do. An alternative is to use unstable theme creator, but it's not recommended in production. TBH, I'm wondering if we should not just leave StrictMode on and get used to this warning - what do you think? 
I'll try to figure out a way of passing a ref to our accordion, but it could be problematic, because the error is thrown by Transition - an element that Accordion is using under the hood.

vilanelle (approved):

src/detail-page/components/accordion/AccordionListItem.tsx
klaraborowska: Probably there is a mistake, cause you want to check, if item has a description, but you are acctually checking url ;)

src/detail-page/components/ResourcesList.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const ResourcesList: React.FC<ResourcesListProps> = ({
+  professionId,
+}) => {
+  const resourcesData = getData(professionId);
+  return (
+    <div>
+      {Object.keys(resourcesData).map((resourceType, index) => {
+        const resourceTitle = ResourcesTypeNames[resourceType as ResourcesType];
+        const areResourcesAvailable = resourcesData[resourceType as ResourcesType].length !== 0;
```
klaraborowska: Just to let you know - there is another way to check, if array is not empty - you could write `!!resourcesData[resourceType as ResourcesType].length`, but of course the way you did it is just fine :)

src/detail-page/components/tabs/TabPanel.tsx
```diff
@@ -9,8 +9,9 @@ interface Props {
 }
 
 const StyledDiv = styled('div')`
-    min-height: 62vh;
-    display: flex;
+  min-height: 62vh;
```
klaraborowska, line 12: I would probably avoid to style height like that, cause there is a risk that is will display wrong, when you change the content above. I would probably add some wrapper for the content and use `flex: 1` to  make the content use whole space that is left, no matter, of the height of items above.

src/detail-page/components/accordion/Accordion.tsx
klaraborowska: yes! great job :)

src/index.tsx
ireshka: I'm afraid that if we leave this `console.error` warning it can be misleading for us as we (or next devs) work in next weeks/months why this error persists.

src/detail-page/components/accordion/AccordionListItem.tsx
ireshka: Ofc :) Thanks for finding it! I corrected it.

src/data/detail-page/resources/ResourcesTypeNames.ts
ireshka: `export type ResourcesType = keyof typeof ResourcesTypeNames` is great for this problem :) I changed the code according to it and I added `enum` instead of `Record`.

src/detail-page/components/LearnTab.tsx
ireshka: Ofc, done.

src/detail-page/components/accordion/Accordion.tsx
ireshka: Ofc, done.

src/detail-page/components/tabs/TabPanel.tsx
ireshka: In my PR I only added some padding here and I'm not familiar enough with the whole styling concept here. Maybe we can move unifying styling tabs to a separate task after adding charts?

klaraborowska: sure, we can do that :)

klaraborowska (approved):

src/index.tsx
vilanelle: The good news is that Material v. 5 will resolve this problem. I'll add a comment that StrictMode should be restored when v. 5 is stable :)

PR #23, opened 2021-05-04
47-empty-search-message

joannajerz:


src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -33,16 +33,22 @@ const SearchSpinner = styled('div')`
...
 const JobTab: React.FC = () => {
   const [role, setRole] = React.useState('');
   const [city, setCity] = React.useState('');
   const [url, setUrl] = React.useState('');
   const handleChangeRole = React.useCallback(e => setRole(e.target.value), []);
   const handleChangeCity = React.useCallback(e => setCity(e.target.value), []);
-  const { jobPosts, isFetching } = useFetchJobs(url);
-
+  /* eslint-disable prefer-const */
+  let { jobPosts, isFetching, emptySearchMessage } = useFetchJobs(url);
```
vilanelle, line 50: We can change it to `const`, we're not reassigning it later :)

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
```
vilanelle, line 1: Personally I'd make the font a bit smaller, but that's probably a matter of opinion :)
image

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -33,16 +33,22 @@ const SearchSpinner = styled('div')`
 `;
 
 const JobList = styled('ul')`
-  padding:0;
+  padding: 0;
 `;
+
+const IsEmpty = styled('div')`
```
vilanelle, line 39: Conventionally, `is` prefix is used for boolean values (you can check out [this short list of most common conventions ](https://hackernoon.com/the-art-of-naming-variables-52f44de00aad) for variable naming). 
This is a component so I'd go for a noun, sth like `EmptySearchMessage`, `NoResultsMessage`.

src/detail-page/components/TabJobContent/JobTab.tsx
```diff
@@ -78,11 +84,12 @@ const JobTab: React.FC = () => {
       >
         Search
       </SearchButton>
-      {jobPosts.length > 0 ? null : <HiringImage src={hiringPicture} />}
       { isFetching ? <SearchSpinner><CircularProgress /></SearchSpinner> : null}
       <JobList>
         {jobPosts.map((job: Job) => <JobListItem job={job} key={job.id} />)}
+        <IsEmpty>{jobPosts.length > 0 ? null : <Typography variant="h2" color="textSecondary"> {emptySearchMessage} </Typography>}</IsEmpty>
```
vilanelle, line 90: I don't think we need to get the empty search message from the hook. The message itself will never change; what we want is to manage it's visibility. 
We can easily achieve it like so:
```suggestion
      {!isFetching && !jobPosts.length && (
          <IsEmpty>
            <Typography variant="h2" color="textSecondary"> No job offers found. </Typography>
          </IsEmpty>
        )}
```
And then we can safely remove all logic related to emptySearchMessage in the hook :)
Finally, we could also extract the condition to a named var:
```
const isEmptySearchMessageVisible = !isFetching && !jobPosts.length;
```

vilanelle (changes requested):

PR #24, opened 2021-05-05
Drm 71 - devops chart

ireshka:
- added `Chart` component as a Container
- added DevOps chart (data + implementation)
- added interfaces & types for charts
end


src/data/detail-page/charts/devOps.ts
```diff
@@ -1,4 +1,61 @@
-import { Chart } from './chartInterfaces';
+import { Chart, ChartDataItem, ChartColor } from './chartInterfaces';
```
vilanelle: I wonder if we should not move the `charts` data folder one level up (from `detail-page` to `data`). Charts are now part of our DetailPage, but they are a nice, separate entity that could be used (or moved) anywhere. What do you think?

src/detail-page/components/charts/DevOpsChart.tsx
```diff
@@ -0,0 +1,51 @@
...
+`;
+
+const StyledSource = styled('p')`
+  font-size: 0.8rem;
+  margin-top: ${({ theme }) => `${theme.spacing(3)}px`};
+  text-align: right;
+`;
+
+// Source Paragraph can be shared between all charts?
+export const DevOpsChart: VFC = () => (
```
vilanelle: I'm always impressed how you manage to find interesting bits and pieces in the docs 👏 In this case, `VFC` seems like a more precise definition for our component, but from what I found it looks like [it's a temporary solution](https://basarat.gitbook.io/typescript/tsx/react#void-function-components), so we should probably go with the usual `FC` instead.

src/detail-page/components/charts/DevOpsChart.tsx
```diff
@@ -0,0 +1,51 @@
...
+  max-width: 400px;
+`;
+
+const StyledSource = styled('p')`
+  font-size: 0.8rem;
+  margin-top: ${({ theme }) => `${theme.spacing(3)}px`};
+  text-align: right;
+`;
+
+// Source Paragraph can be shared between all charts?
```
vilanelle: Definitely, it would be great if we had a single source component for source labels 💪

src/detail-page/components/LearnTab.tsx
```diff
@@ -10,15 +11,21 @@ const TabContainer = styled('div')`
   width: 100%;
 `;
 
+const PROFESSION_CHARTS = {
+  frontend: null,
+  backend: null,
+  uxui: null,
+  qa: null,
+  devops: <DevOpsChart />,
+  projectmanager: null,
```
vilanelle, line 20: I know it's not really part of your PR, but flat case is not really used in JS. Maybe we could change `uxui`, `devops` and `projectmanager` to camelCase? (`uxUi`, `devOps`, `projectManager`).

src/data/detail-page/charts/chartInterfaces.ts
```diff
@@ -1,4 +1,17 @@
 export interface Chart {
-    data?: string[];
-    type?: string;
+  data?: ChartDataItem[];
```
vilanelle: We have multiple optional properties on this interface and this is not a good sign - if we make everything optional we're giving up much of the protection we get from TypeScript. 
Let's reconsider this interface: it's unlikely that we will ever have a chart without `data` or `type` so we can probably make them both non-optional. For the others, we can make them `string | null` or `Source | null` and remove the optional `?`. That way we will have to explicitly set them to null if we want to skip them (if ever).

src/data/detail-page/charts/devOps.ts
```diff
@@ -1,4 +1,61 @@
...
 
-export const devOps: Chart = {
+interface DevOpsChartDataItem extends ChartDataItem {
+  color: ChartColor;
+}
+
+type DevOpsChart = Omit<Chart, 'data'> & { data: DevOpsChartDataItem[]};
```
vilanelle: I'm not sure about substituting data type for a given chart. In this case, I'd rather go for optional `color` prop on `ChartDataItem` and handling potential override in the place where we use the data.

vilanelle (changes requested):

src/detail-page/components/charts/DevOpsChart.tsx
```diff
@@ -0,0 +1,70 @@
...
+      ...(labelText && {
+        label: labelText,
+      }),
+      ...(areColorsDefined && {
+        backgroundColor: chartBackgroundColors,
+        borderColor: chartBorderColors,
+      }),
+    },
+  ],
+};
```
ireshka: It's definitely not the best way to do that but chart.js types required not null/undefined value for any datasets key so I have to filter/cast it.

src/detail-page/components/charts/Chart.tsx
```diff
@@ -0,0 +1,47 @@
...
+  text-align: right;
+`;
+
+export const Chart: FC<ChartInterface> = ({ children, maxWidth, title, source, type }) => {
+  const chartId = `chart-${type}`;
+  return (
+    <StyledWrapper>
+      <StyledTitle variant="subtitle2" align="center" id={chartId}>
+        {title}
+      </StyledTitle>
```
ireshka: I decided to add title as a separate element because of accessibility issues - https://github.com/reactchartjs/react-chartjs-2/issues/373 and I think we have more control about it (e.g. styling issues).

src/detail-page/components/charts/Chart.tsx
```diff
@@ -0,0 +1,47 @@
...
+  margin-bottom: ${({ theme }) => `${theme.spacing(2)}px`};
+`;
+
+const StyledSource = styled('p')`
+  font-size: 0.8rem;
+  margin-top: ${({ theme }) => `${theme.spacing(3)}px`};
+  text-align: right;
+`;
+
+export const Chart: FC<ChartInterface> = ({ children, maxWidth, title, source, type }) => {
```
ireshka: I create component that can share title/source and other element with all charts.

src/detail-page/components/charts/DevOpsChart.tsx
```diff
@@ -0,0 +1,70 @@
...
+        const tooltipIndex = tooltipItem.index!;
+        const tooltipLabel = datasets![tooltipDatasetIndex].label || '';
+        const tooltipValue = datasets![tooltipDatasetIndex].data![tooltipIndex];
+        const valueLabel = labels![tooltipIndex];
+        return ` ${valueLabel}: ${tooltipValue} ${tooltipLabel}`;
+      },
+    },
+  },
+};
+
```
ireshka: I remember that @klaraborowska talked about her experience with tooltips so I'm open to serious refactor this solution if needed!

src/data/detail-page/charts/devOps.ts
ireshka: I moved it one level up. Good idea.

src/detail-page/components/charts/DevOpsChart.tsx
ireshka: In the next major release there will be probably a change with `FC` component too  :smile: but because we didn't use `VFC` anywhere else you're right that we should get back to `FC`.

src/detail-page/components/charts/DevOpsChart.tsx
ireshka: I expanded idea and created another component for common chart parts.

src/data/detail-page/charts/chartInterfaces.ts
ireshka: Chart interface is now slightly refactored.

src/data/detail-page/charts/devOps.ts
ireshka: Done

src/detail-page/components/charts/DevOpsChart.tsx
ireshka: azure

package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@material-ui/core": "^4.11.2",
     "@material-ui/icons": "^4.11.2",
+    "@types/chart.js": "^2.9.32",
```
vilanelle, line 8: :+1:

src/detail-page/components/charts/Chart.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { FC } from 'react';
+import styled from 'styled-components';
+import { Typography } from '@material-ui/core';
+import { ChartSource } from '../../../data/charts/chartInterfaces';
+import { ProfessionId } from '../../../data/home-page/ITProfessions';
+
+interface ChartInterface {
```
vilanelle: It's a small thing, but I'd go with `Props` or `ChildProps` to keep consistency with the rest of the app.

src/detail-page/components/charts/Chart.tsx
vilanelle: I like it a lot! :clap: Also, maybe we could reflect that fact that it's a wrapper in the name and call it e.g. `ChartWrapper`? It doesn't really have any chart logic, because we pass the chart as child :)

src/detail-page/components/charts/DevOpsChart.tsx
```diff
@@ -0,0 +1,70 @@
...
+  tooltips: {
+    callbacks: {
+      label(tooltipItem, data) {
+        const { datasets, labels } = data;
+        const tooltipDatasetIndex = tooltipItem.datasetIndex!;
+        const tooltipIndex = tooltipItem.index!;
+        const tooltipLabel = datasets![tooltipDatasetIndex].label || '';
+        const tooltipValue = datasets![tooltipDatasetIndex].data![tooltipIndex];
+        const valueLabel = labels![tooltipIndex];
+        return ` ${valueLabel}: ${tooltipValue} ${tooltipLabel}`;
```
vilanelle: [very picky] Let's remove the space between tooltip value and label ;)

vilanelle (approved):

src/detail-page/components/charts/Chart.tsx
ireshka: Done.

src/detail-page/components/charts/Chart.tsx
ireshka: You're right. I mixed convention from another project :sweat_smile:

src/detail-page/components/charts/DevOpsChart.tsx
```diff
@@ -0,0 +1,70 @@
...
+const ChartData: Chart.ChartData = {
+  labels: chartLabels,
+  datasets: [
+    {
+      data: chartValues,
+      borderWidth: 1,
+      ...(labelText && {
+        label: labelText,
+      }),
+      ...(areColorsDefined && {
```
klaraborowska: I'm not sure about this condition. I understand the one above with labelText, as it may happen that we don't want to display labels. But probably we do want to have colors for any type that we have in our pie chart. Or maybe I missed something here? :wink:  The same thing with filtering the colors array. I'm afraid that with filtering out empty color values, we could unintentionally hide the fact that the color for one of our data type is missing.  What do you think about it? :blush: Maybe I just didn't notice/misunderstood something?

src/detail-page/components/charts/DevOpsChart.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { chartColors } from '../../../data/charts/chartColors';
+import { ChartWrapper } from './ChartWrapper';
+import { notEmpty } from '../../../shared/utils/helpers';
+
+const { data: sourceData, labelText, source, title: chartTitle, type } = devOps;
+
+const chartLabels = sourceData.map(({ label }) => label);
+const chartValues = sourceData.map(({ value }) => value);
+const chartBackgroundColors = sourceData
+  .map(({ color }) => (color ? chartColors[color].light : null))
```
klaraborowska: We could get rid of one pair of parenthesis here :blush: `.map(({ color }) => color ? chartColors[color].light : null)` would be enough

src/detail-page/components/charts/DevOpsChart.tsx
ireshka: `areColorsDefined` - this is is check if one of our color is not missing. If there is at least one null - colors won't be added because one of their value will be null or undefined - `chart.js` allowed only string as value of color.
filtering not empty is for type guarding - because without it `chartBorderColors` has union type `(null | string)[]` & only `string[]` is allowed (again chart.js types).
Maping with color/null is because of optional property `color` in `ChartItemData` interface.

src/detail-page/components/charts/DevOpsChart.tsx
ireshka: I've corrected it. My prettier sometimes has strange ideas ;)

src/detail-page/components/charts/DevOpsChart.tsx
klaraborowska: Gosia I understand your point :)  my question was more like - is there any situation, when we want to display our chart without colors? maybe we should just change our typings and make color requiered, not optional?

ireshka:
Following our discussion, I changed `ChartDataItem` interface.

src/data/charts/chartInterfaces.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export interface ChartSource {
+  link: string;
+  description: string;
+}
+
+export interface ChartDataItem {
+  label: string;
+  value: number;
+  color: ChartColor;
```
ireshka, line 19: Now color is mandatory.

src/detail-page/components/charts/DevOpsChart.tsx
```diff
@@ -0,0 +1,62 @@
...
+      data: chartValues,
+      borderWidth: 1,
+      ...(labelText && {
+        label: labelText,
+      }),
+      backgroundColor: chartBackgroundColors,
+      borderColor: chartBorderColors,
+    },
+  ],
+};
```
ireshka, line 28: I've removed excess checking/filtering.

